PR #157, opened 2024-04-12
Improve Python E2E tests and align with Java.

zzhlogin:
*Issue #, if available:*

Improve Python E2E tests with following changes and align with Java:
1. Run E2E in docker contianer, matching with Java PR: https://github.com/aws-observability/aws-application-signals-test-framework/pull/36
2. Append testing id to sample-app-namespace, matching with Java PR: https://github.com/aws-observability/aws-application-signals-test-framework/pull/38

Test workflow run:
1. EC2:https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/8666582644
2. EKS: https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/8667061312

### DIFF
--- a/.github/workflows/application-signals-python-e2e-ec2-test.yml
+++ b/.github/workflows/application-signals-python-e2e-ec2-test.yml
@@ -32,11 +32,13 @@ env:
   METRIC_NAMESPACE: AppSignals
   LOG_GROUP_NAME: /aws/appsignals/generic
   ADOT_WHEEL_NAME: ${{ inputs.staging_wheel_name }}
-  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
+#  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
 
 jobs:
   python-e2e-ec2-test:
     runs-on: ubuntu-latest
+    container:
+      image: public.ecr.aws/h6o3z5z9/aws-application-signals-test-framework-workflow-container:latest
     steps:
       - name: Get testing resources from aws-application-signals-test-framework
         uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


harrryr (approved):

.github/workflows/application-signals-python-e2e-ec2-test.yml
```diff
@@ -32,11 +32,12 @@ env:
   METRIC_NAMESPACE: AppSignals
   LOG_GROUP_NAME: /aws/appsignals/generic
   ADOT_WHEEL_NAME: ${{ inputs.staging_wheel_name }}
-  TEST_RESOURCES_FOLDER: /home/runner/work/aws-application-signals-test-framework/aws-application-signals-test-framework
```
majanjua-amzn, line 35: Out of curiosity, why is this not needed anymore?

.github/workflows/application-signals-python-e2e-ec2-test.yml
```diff
@@ -32,11 +32,12 @@ env:
...
 
 jobs:
   python-e2e-ec2-test:
     runs-on: ubuntu-latest
+    container:
+      image: public.ecr.aws/h6o3z5z9/aws-application-signals-test-framework-workflow-container:latest
```
majanjua-amzn, line 40: Are there any tools/packages that the python test uses that can be added to the Dockerfile we created in the test framework?

majanjua-amzn (approved):

.github/workflows/application-signals-python-e2e-ec2-test.yml
zzhlogin: Since this is the adot python repo, we did full github checkout for the `aws-application-signals-test-framework` and use the codes there, so don't need to use the full path. I will need to add it back when moving E2E tests to test-framework repo.

.github/workflows/application-signals-python-e2e-ec2-test.yml
zzhlogin: There is no extra need for python. The E2E test steps are very similar as Java, all the python related installation are added through terraform instead of workflow.

PR #158, opened 2024-04-12
Attempt to fix test_urls_excluded_from_sampling

thpierce:
`test_urls_excluded_from_sampling` keeps failing on PR, causing annoying delays. I suspect that the problem is a race condition - we are exporting a span then grabbing it, and I believe that the delay between export and being "grabbable" is causing the test to fail (typically with `1 != 2`). As an attempted fix, I'm adding some retry logic that will attempt to grab for a full second.


jj22ee (approved):
LGTM

In the future, we will move away from using ENV vars to ignore select URLs, use upstream `suppress_instrumentation()` method instead

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2518a4ac07cb62ad6587dd8f6cbb5f8663a7e179/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py#L189-L192

PR #162, opened 2024-04-15
Post release 0.1.0: Update version to 0.1.0.dev0

srprash:
This PR prepares the main branch for the next development cycle by updating the version to 0.1.0.dev0 and updating the image version to be scanned to the latest released.
          
This PR should only be merge when release for version v0.1.0 is success.

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           image_uri_with_tag: "adot-autoinstrumentation-python:test"
           push_image: false
-          load_image: false
+          load_image: true
           python_version: "3.10"
           package_name: aws-opentelemetry-distro
           os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/release_build.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           image_uri_with_tag: "adot-autoinstrumentation-python:test"
           push_image: false
-          load_image: false
+          load_image: true
```
srprash, line 35: This change comes from the fix that we did on the `release/v0.1.x` branch: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/160

XinRanZhAWS: was about to ask, thanks for call out

thpierce: Lets cut a backlog item that asserts that either `push_image` or `load_image` is true.

XinRanZhAWS (commented):
There is a check fail, should we record and rerun?

thpierce (approved):

XinRanZhAWS (approved):

.github/workflows/release_build.yml
srprash: done

PR #179, opened 2024-05-07
insert CWD to sys.path for Django to find the application directory

srprash:
*Issue #, if available:* https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2495
The issue is that users on K8s/EKS/ECS explicitly need to set their Django application container's working directory in the `PYTHONPATH` environment variable ([operator code](https://github.com/open-telemetry/opentelemetry-operator/blob/d42a0ce1166efe86c95f85268f265e338d0002f8/pkg/instrumentation/python.go#L55-L63) that handles the `PYTHONPATH`), otherwise the application fails to start with error messages related to the Django's settings module.

*Analysis:*
Python auto-instrumentation using the `opentelemetry-instrument` command doesn't have this issue, which is I believe is because of [this specific handling of Django app current directory](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/1ee7261ea7117fbd22e2262e488402213a874125/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py#L98-L102) in the auto-instrumentation startup. Since [this script is only run when using the `opentelemetry-instrument` command](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/1ee7261ea7117fbd22e2262e488402213a874125/opentelemetry-instrumentation/pyproject.toml#L34), auto-instrumentation done by the operator (using the [sitecustomize](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py) I believe) doesn't set the current working directory in the PYTHONPATH.

*Description of changes:*
Solution is to insert the current working directory to the `sys.path` within the auto-instrumentation so that Django can find the application folder. We are adding to the `sys.path` and not `PYTHONPATH` because at this point the python process has already started and mutations to `PYTHONPATH` doesn't reflect in the `sys.path` which is what Django uses.

*Testing:*
Manually tested by deploying a [Django sample application](https://github.com/srprash/otel-python-k8s-samples/tree/main/django) to minikube without setting the `PYTHONPATH` in the container environment.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
@@ -1,6 +1,8 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 import os
+import sys
+from logging import Logger, getLogger
 
 from amazon.opentelemetry.distro.patches._instrumentation_patch import apply_instrumentation_patches
 from opentelemetry.distro import OpenTelemetryDistro
@@ -9,6 +11,8 @@ from opentelemetry.sdk.environment_variables import (
     OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION,
     OTEL_EXPORTER_OTLP_PROTOCOL,
 )
+
+_logger: Logger = getLogger(__name__)
 
 
 class AwsOpenTelemetryDistro(OpenTelemetryDistro):
@@ -36,6 +40,21 @@ class AwsOpenTelemetryDistro(OpenTelemetryDistro):
             OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION environment variable. Need to work with upstream to
             make it to be configurable.
         """
+
+        # related issue: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2495
+        # mimicking what is done here: https://tinyurl.com/54mvzmte
+        # For handling applications like django running in containers, we are setting the current working directory
+        # to the sys.path for the django application to find its executables.
+        #
+        # Note that we are updating the sys.path and not the PYTHONPATH env var, because once sys.path is
+        # loaded upon process start, it doesn't refresh from the PYTHONPATH value.
+        #
+        # To be removed once the change has been contributed to upstream.
+        cwd_path = os.getcwd()
+        _logger.debug("Current working directory path: %s", cwd_path)
+        if cwd_path not in sys.path:
+            sys.path.insert(0, cwd_path)
+
         os.environ.setdefault(OTEL_EXPORTER_OTLP_PROTOCOL, "http/protobuf")
 
         super(AwsOpenTelemetryDistro, self)._configure()
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
@@ -41,7 +41,7 @@ class AwsOpenTelemetryDistro(OpenTelemetryDistro):
             make it to be configurable.
         """
 
-        # related issue: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2495
+        # Issue: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2495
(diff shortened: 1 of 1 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
```diff
@@ -36,6 +40,21 @@ def _configure(self, **kwargs):
...
             make it to be configurable.
         """
+
+        # related issue: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2495
+        # mimicking what is done here: https://tinyurl.com/54mvzmte
+        # For handling applications like django running in containers, we are setting the current working directory
+        # to the sys.path for the django application to find its executables.
+        #
+        # Note that we are updating the sys.path and not the PYTHONPATH env var, because once sys.path is
+        # loaded upon process start, it doesn't refresh from the PYTHONPATH value.
```
thpierce, line 50: Not totally sure I understand this - why does the upstream [change](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/0a231e57f9722e6101194c6b38695addf23ab950/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py#L98-L102) not have to worry about this?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
```diff
@@ -36,6 +40,21 @@ def _configure(self, **kwargs):
...
+        #
+        # To be removed once the change has been contributed to upstream.
```
thpierce, line 52: Did we raise an upstream PR? If so can we link it here?

Also lets avoid using the term upstream, it's vague. Lets specify the repo or class that needs this change.

thpierce (commented):
Not blocking, but a question: Does this work if the docker `CMD` running the application to be instrumented is not in the same dir as the application? E.g. `python ./run.py` vs `python ./subdir/run.py` or the like?

srprash:
>Does this work if the docker CMD running the application to be instrumented is not in the same dir as the application? E.g. python ./run.py vs python ./subdir/run.py or the like?

@thpierce Unfortunately, this solution will not work when Django application is run from within a subdirectory without changing into that directory. I ran a quick test with the following Dockerfile and it did not work. But this is limitation we can't solve right now.
```
FROM python:3.11

WORKDIR /app
RUN mkdir -p subdir

# Install dependencies
COPY djangoApp/requirements.txt /app/subdir

RUN pip install -r subdir/requirements.txt

# Copy the project code into the container
COPY djangoApp/. /app/subdir

EXPOSE 8000

CMD ["python", "subdir/manage.py", "runserver", "0.0.0.0:8000", "--noreload"]
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
srprash: Did some more digging to understand why the Otel code works by s[etting the CWD in the ](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py#L98-L102)`PYTHONPATH` but in ADOT we have to set the CWD in the `sys.path`.
The full command to run auto-instrumentation for on-prem applications is `opentelemetry-instrument python app.py`. Here, the `opentelemetry-instrument` runs first, sets the CWD in `PYTHONPATH` and then the `python app.py` invokes a new python process. When the new process starts, the `PYTHONPATH` entries are populated in the `sys.path`.
When not using the `opentelemetry-instrument` command (such as in K8s) the auto-instrumentation starts with [sitecustomize.py](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py) as the entry-point, the `python app.py` process has already started and the `sys.path` has already read the `PYTHONPATH`. So we need to update the `sys.path` directly as `PYTHONPATH` mutation has no effect on the `sys.path` within the same process.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
srprash: I did't raise a PR in OTel yet. Will do so soon and link it here.

thpierce (approved):

PR #182, opened 2024-05-14
Apply DB resource attributes.

zzhlogin:
Apply DB resource attributes: `RemoteResourceType` and `RemoteResourceIdentifier` with the following details.
This PR have feature parity with adot Java: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/805 

The only difference it python does not allow same function with different number of inputs as Java, So I'm renaming the second `_build_db_connection` to `_build_db_connection_string`.

Tested in EC2 and see correct trace, metric, service graphs generated: 

Screenshot 2024-05-16 at 3 00 57 PM
Screenshot 2024-05-16 at 3 01 23 PM

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
@@ -409,11 +409,11 @@ def _get_db_connection(span: ReadableSpan) -> None:
         db_connection = _build_db_connection(server_address, server_port)
     elif is_key_present(span, _NET_PEER_NAME):
         network_peer_address: Optional[str] = span.attributes.get(_NET_PEER_NAME)
-        network_peer_port: Optional[str] = span.attributes.get(_NET_PEER_PORT)
+        network_peer_port: Optional[int] = span.attributes.get(_NET_PEER_PORT)
         db_connection = _build_db_connection(network_peer_address, network_peer_port)
     elif is_key_present(span, _SERVER_SOCKET_ADDRESS):
         server_socket_address: Optional[str] = span.attributes.get(_SERVER_SOCKET_ADDRESS)
-        server_socket_port: Optional[str] = span.attributes.get(_SERVER_SOCKET_PORT)
+        server_socket_port: Optional[int] = span.attributes.get(_SERVER_SOCKET_PORT)
         db_connection = _build_db_connection(server_socket_address, server_socket_port)
     elif is_key_present(span, _DB_CONNECTION_STRING):
         connection_string: Optional[str] = span.attributes.get(_DB_CONNECTION_STRING)
@@ -430,9 +430,10 @@ def _build_db_connection(address: str, port: int) -> Optional[str]:
 
 
 def _build_db_connection_string(connection_string: str) -> Optional[str]:
+
+    uri = urlparse(connection_string)
+    address = uri.hostname
     try:
-        uri = urlparse(connection_string)
-        address = uri.hostname
         port = uri.port
     except ValueError:
         return None
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
@@ -379,6 +379,9 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
 
 def _get_db_remote_resource_identifier(span: ReadableSpan) -> str:
     db_remote_resource_identifier: str = ""
+    print("SPAN!!!!!!!!!!!!!!!")
+    print(span)
+    print("================================")
     if is_key_present(span, _DB_NAME):
         db_remote_resource_identifier += span.attributes.get(_DB_NAME) + "|"
 
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
@@ -357,8 +357,9 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
         and not is_key_present(span, _RPC_SERVICE)
         and not is_key_present(span, _RPC_METHOD)
     ):
-        remote_resource_type = "DB::Endpoint"
         remote_resource_identifier = _get_db_remote_resource_identifier(span)
+        if remote_resource_identifier:
+            remote_resource_type = "DB::Endpoint"
     elif is_key_present(span, AWS_STREAM_NAME):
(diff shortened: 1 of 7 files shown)


contract-tests/tests/test/amazon/botocore/botocore_test.py
```diff
@@ -176,8 +176,6 @@ def test_dynamodb_create_table(self):
             0,
             remote_service="AWS::DynamoDB",
             remote_operation="CreateTable",
-            remote_resource_type="AWS::DynamoDB::Table",
```
thpierce, line 179: I don't think this is expected behaviour. AWS SDK remote resource info should not be impacted.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -344,10 +348,18 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
     remote_resource_type: Optional[str] = None
     remote_resource_identifier: Optional[str] = None
 
-    # Only extract the table name when _AWS_TABLE_NAMES has size equals to one
-    if is_key_present(span, _AWS_TABLE_NAMES) and len(span.attributes.get(_AWS_TABLE_NAMES)) == 1:
-        remote_resource_type = _NORMALIZED_DYNAMO_DB_SERVICE_NAME + "::Table"
-        remote_resource_identifier = span.attributes.get(_AWS_TABLE_NAMES)[0]
+    # Only extract remote_resource_type and remote_resource_identifier
```
thpierce, line 351: I think we should consider tying the implementation of this method to `_set_remote_service_and_operation`. E.g. Change `_set_remote_service_and_operation` to `_set_remote_attributes` and have it set remote service, operation, resource type, and resource identifier all in one go. All attributes are tied together, so it makes sense to have the logic be centralized.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -344,28 +353,100 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
...
+    - {SpanAttributes.DB_CONNECTION_STRING}-Port
+
+    If address is not present, neither RemoteResourceType nor RemoteResourceIdentifier will be provided.
+    """
+    db_name: Optional[str] = span.attributes.get(_DB_NAME)
+    db_connection: Optional[str] = None
+
+    if is_key_present(span, _SERVER_ADDRESS):
+        server_address: Optional[str] = span.attributes.get(_SERVER_ADDRESS)
+        server_port: Optional[str] = span.attributes.get(_SERVER_PORT)
```
thpierce, line 408: Is this actually a string and not a long/int? Ditto to others.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -344,28 +353,100 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
...
+
+def _build_db_connection_string(connection_string: str) -> Optional[str]:
+    uri = urlparse(connection_string)
+    address = uri.hostname
+    port = uri.port
+
+    if address is None:
+        return None
+
+    port_str = "|" + str(port) if port is not None and port != -1 else ""
```
thpierce, line 440: Java and python do not handle invalid ports the same here: https://docs.python.org/3/library/urllib.parse.html

> Changed in version 3.6: Out-of-range port numbers now raise [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError), instead of returning [None](https://docs.python.org/3/library/constants.html#None).

Need to wrap in try/catch.

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_aws_metric_attribute_generator.py
```diff
@@ -911,14 +911,23 @@ def _validate_peer_service_does_override(self, remote_service_key: str) -> None:
...
+        ]
+        values: List[str] = [
+            "aws-api",
+        ]
+        self._mock_attribute(keys, values)
         # Validate behaviour of aws bucket name attribute, then remove it.
-        self._mock_attribute([SpanAttributes.AWS_S3_BUCKET], ["aws_s3_bucket_name"])
+        self._mock_attribute(
+            [SpanAttributes.AWS_S3_BUCKET, SpanAttributes.RPC_SYSTEM], ["aws_s3_bucket_name", "aws-api"], keys, values
+        )
```
thpierce, line 925: Why do you need to do this here, but for no other test? E.g. why not just:
```
 self._mock_attribute([SpanAttributes.AWS_S3_BUCKET], ["aws_s3_bucket_name"], keys, values)
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
zzhlogin: updated.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
thpierce: Ok others should be updated, but this isn't blocking.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -430,9 +430,12 @@ def _build_db_connection(address: str, port: int) -> Optional[str]:
...
 def _build_db_connection_string(connection_string: str) -> Optional[str]:
-    uri = urlparse(connection_string)
-    address = uri.hostname
-    port = uri.port
+    try:
+        uri = urlparse(connection_string)
+        address = uri.hostname
+        port = uri.port
+    except ValueError:
+        return None
```
thpierce, line 438: This is wrong. Fix:

```
    uri = urlparse(connection_string)
    address = uri.hostname
    try:
        port = uri.port
    except ValueError:
        port = None
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -405,7 +405,7 @@ def _get_db_connection(span: ReadableSpan) -> None:
 
     if is_key_present(span, _SERVER_ADDRESS):
         server_address: Optional[str] = span.attributes.get(_SERVER_ADDRESS)
-        server_port: Optional[str] = span.attributes.get(_SERVER_PORT)
+        server_port: Optional[int] = span.attributes.get(_SERVER_PORT)
```
thpierce, line 408: Not blocking: Should be done of other if/else blocks.

thpierce (approved):

PR #195, opened 2024-05-24
Fail artifact build if load and push are both false

thpierce:
If both are false, we get somewhat arcane failures. All existing workflows set one to true and the other to false, this just enforces it.

Also, Fix a codespell issue that was introduced in 2.3.0 (previously using 2.2.6).

Testing: Set  both options to false in pr_build- https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/9230011064/job/25397368739?pr=195


srprash (approved):

vastin (approved):
LGTM

PR #199, opened 2024-06-03
Update requests version used in contract tests

thpierce:
GitHub complains about relying on 2.31.0 so changing this to pull the latest 2.x version, which is actually https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation says it supports: `requests ~= 2.0`


vastin (approved):
LGTM. ty.

PR #200, opened 2024-06-03
Enhance kinesis Consumer support

zzhlogin:
Add kinesis Consumer Support with the following change:

1. Modify patch `_KinesisExtension` where extract `ConsumerARN` and add into `"aws.kinesis.consumer_arn"` span attributes. The attributes will be retrieved from both request and response context.

2. Populate `RemoteResourceType` and `RemoteResourceIdentifier` for SNS in ADOT python and update unit test and contract test to verify ADOT python performance.

3. Contract test - `botocore_server.py`: As we only want to have one span per test, and the consumerArn contains random prefix, it will be different per test, we pre-compute `consumer_arn` and pass it into RequestHandler. To avoid exception when creating resource, we check if the resource exist first, and create the resource when it not found. This make sure `prepare_aws_server` function code can be run to the end without exception and the `consumer_arn` can be returned.

3. Contract test - `contract_test_base.py`: As `ConsumerARN` can include random prefix, we add `_assert_match_attribute` function to check if the identifier match with certain pattern. 

Testing:
A manually E2E test is performed, and confirmed the expected traces metrics and service maps are generated:

Trace: 
Screenshot 2024-06-03 at 5 22 15 PM

Metrics:
Screenshot 2024-06-03 at 5 23 47 PM

Service Map:

Screenshot 2024-06-03 at 5 24 03 PM


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_attribute_keys.py
```diff
@@ -16,3 +16,4 @@
 AWS_QUEUE_URL: str = "aws.sqs.queue_url"
 AWS_QUEUE_NAME: str = "aws.sqs.queue_name"
 AWS_STREAM_NAME: str = "aws.kinesis.stream_name"
+AWS_CONSUMER_ARN: str = "aws.kinesis.consumer_arn"
```
thpierce, line 19: Maybe more clear:
```suggestion
AWS_STREAM_CONSUMER_ARN: str = "aws.kinesis.stream_consumer_arn"
```

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_aws_metric_attribute_generator.py
```diff
@@ -950,6 +951,19 @@ def test_sdk_client_span_with_remote_resource_attributes(self):
         self._validate_remote_resource_attributes("AWS::Kinesis::Stream", "aws_stream_name")
         self._mock_attribute([AWS_STREAM_NAME], [None])
 
+        # Validate behaviour of AWS_CONSUMER_ARN attribute, then remove it.
```
thpierce, line 954: Add a test covering the case when both stream name and consumer ARN are present.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
```diff
@@ -361,6 +362,9 @@ def _set_remote_type_and_identifier(span: ReadableSpan, attributes: BoundedAttri
         elif is_key_present(span, AWS_STREAM_NAME):
             remote_resource_type = _NORMALIZED_KINESIS_SERVICE_NAME + "::Stream"
             remote_resource_identifier = _escape_delimiters(span.attributes.get(AWS_STREAM_NAME))
+        elif is_key_present(span, AWS_CONSUMER_ARN):
+            remote_resource_type = _NORMALIZED_KINESIS_SERVICE_NAME + "::StreamConsumer"
+            remote_resource_identifier = _escape_delimiters(span.attributes.get(AWS_CONSUMER_ARN))
```
thpierce, line 367: Is it possible that both stream name and consumer arn are present? Which is the right one to use in that case, why?

contract-tests/images/applications/botocore/botocore_server.py
```diff
@@ -212,17 +220,24 @@ def set_main_status(status: int) -> None:
...
+        if put_bucket_name not in bucket_names:
+            s3_client.create_bucket(
+                Bucket=put_bucket_name, CreateBucketConfiguration={"LocationConstraint": _AWS_REGION}
+            )
+
+        get_bucket_name: str = "test-get-object-bucket-name"
+        if get_bucket_name not in bucket_names:
+            s3_client.create_bucket(
+                Bucket=get_bucket_name, CreateBucketConfiguration={"LocationConstraint": _AWS_REGION}
+            )
```
thpierce, line 240: Why is this needed? Same with changes to DDB, SQS sections? Is it because we are not refreshing the AWS container between tests?

contract-tests/images/applications/botocore/botocore_server.py
```diff
@@ -231,31 +246,54 @@ def prepare_aws_server() -> None:
...
+
+        return [consumer_arn]
     except Exception as exception:
         print("Unexpected exception occurred", exception)
+        return [None]
 
 
 def main() -> None:
-    prepare_aws_server()
+    [consumer_arn] = prepare_aws_server()
```
thpierce, line 291: Suggestion: If `consumer_arn` is None, log some message and fail the container.

contract-tests/images/applications/botocore/botocore_server.py
zzhlogin: For tests, we will pull the "aws-application-signals-tests-botocore-app", and start a new container. Only the first container will succeed, all the following containers are fail with `CreateBucket`, exception: 
`[INFO] Unexpected exception occurred An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it.`
Then the function directly return `[None]` without proceed with following code.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_attribute_keys.py
zzhlogin: Updated.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py
zzhlogin: I checked all APIs that uses consumerARN, there is no case that both stream name and consumer arn are present. So we should be safe in this case. But it is a good callout. We need to double check to other upcoming services.

contract-tests/images/applications/botocore/botocore_server.py
zzhlogin: For the other test cases except `test_kinesis_describe_stream_consumer` , the test can succeed even `consumer_arn` is None as it will not be used. Does it make sense that we let `None` pass into RequestHandler, and use the API call to failed the container/test?

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_aws_metric_attribute_generator.py
zzhlogin: Added.

contract-tests/images/applications/botocore/botocore_server.py
thpierce: Interesting, thanks for investigating and improving!

zzhlogin: Added.

thpierce:
Very outdated, resolve.

PR #201, opened 2024-06-04
Fix Flaky Test in AWS X-Ray Remote Sampler Client Test

jj22ee:
*Issue #, if available:*
This PR fixes flaky test `test_urls_excluded_from_sampling`, causing issues in PR workflows
- Example: https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/9103811387/job/25026448867
 
Instantiation of `AwsXRayRemoteSampler` will spawn processes that will regularly (every 10s and 300s) make a Request call, and these processes stay alive between different unit tests. This means that the test that counts the number instrumented Request calls is affected by these "orphaned" processes.

The solution is to stop these processes inside the tests that they are created from, or stop them from being spawned in the first place if they are not needed.

*Description of changes:*
- Stop the daemon rule and target poller processes inside the tests that they are created from, or stop them from being spawned in the first place if they are not needed.
- Revert previous fix attempt
- Add test for version to increase and pass code coverage

*Testing:*
- Reproduced failing test by running test in loop of 10000 runs
- Failing test not reproducible after applying fix.

### DIFF
--- a/aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/sampler/test_aws_xray_remote_sampler.py
+++ b/aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/sampler/test_aws_xray_remote_sampler.py
@@ -49,96 +49,93 @@ def mocked_requests_get(*args, **kwargs):
 
 
 class TestAwsXRayRemoteSampler(TestCase):
-    def test_create_remote_sampler_with_empty_resource(self):
-        rs = AwsXRayRemoteSampler(resource=Resource.get_empty())
-        self.assertIsNotNone(rs._rules_timer)
-        self.assertEqual(rs._AwsXRayRemoteSampler__polling_interval, 300)
-        self.assertIsNotNone(rs._AwsXRayRemoteSampler__xray_client)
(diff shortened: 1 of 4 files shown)


aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/sampler/test_aws_xray_remote_sampler.py
```diff
@@ -57,6 +57,10 @@ def test_create_remote_sampler_with_empty_resource(self):
         self.assertIsNotNone(rs._AwsXRayRemoteSampler__resource)
         self.assertTrue(len(rs._AwsXRayRemoteSampler__client_id), 24)
 
+        # Clean up timers
+        rs._rules_timer.cancel()
+        rs._targets_timer.cancel()
+
```
thpierce, line 63: Can we add a `setUp` and `tearDown` methods to this class to ensure we do not add flaky tests in future? Can make `rs` a class variable I think?

jj22ee: Added your suggestion in latest commit

thpierce (approved):

PR #210, opened 2024-06-14
Change CodeQL to advanced

thpierce:
The default CodeQL configuration does not run for forked PRs, causing PRs to simply get stuck. In this PR I have enabled "Advanced configuration", and made no changes to the templated workflow that was created. I tested this configuration in https://github.com/thpierce/aws-otel-python-instrumentation/actions, and it seems to behave the same as default configuration, but runs for forked PRs. This is a 2-way door, we can always go back to default if this proves to be the wrong approach.


vastin (approved):
LGTM. ty.

PR #214, opened 2024-06-19
Replace Dockerfile with Scratch as Base Image

harrryr:
*Issue #, if available:*
Currently amazon linux is used as the base image for OTEL python distribution. This image contains a lot of unnecessary packages that inflate the image and also cause issues during high sev scans. Ideally we would like to use an image with minimal content.

`Scratch` was the base image that was proposed to be used with `cpUtility` installed to enable the `cp` command to copy the python image to the sample app pod. The [cpUtility](https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/tools/cp-utility) is copied from aws-otel-java-instrumentation repo, but there is a small issue where if copied from a folder, the destination directory is not correct. 

This piece of [code](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/tools/cp-utility/src/main.rs#L79-L87) in cp-utility causes the directory to change if the destination folder already exists. It was [modified](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/214/files#diff-255e1973e4371a91e310ea9c1b94c1507bab752b6aa20b39322542e6d205616aR77-R80) so that that if the destination folder exists, just reuse it and don't create a new folder. Other than also modifying the [unit test](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/214/files#diff-255e1973e4371a91e310ea9c1b94c1507bab752b6aa20b39322542e6d205616aR269-R308), the code is identical to the otel java

Test run: https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/9586337462

Also manually built the image locally and deployed to EKS cluster with sample app. Verified that correct telemetry was generated and `opentelemetry` folder is located in correct path

Additionally, ran testing in ARM64 EC2 instance


Dockerfile
```diff
@@ -3,6 +3,8 @@
 #   The packages are installed in the `/autoinstrumentation` directory. This is required as when instrumenting the pod by CWOperator,
 #   one init container will be created to copy all the content in `/autoinstrumentation` directory to app's container. Then
 #   update the `PYTHONPATH` environment variable accordingly. Then in the second stage, copy the directory to `/autoinstrumentation`.
+
+# Stage 1: Build the /autoinstrumentation folder
```
srprash, line 7: Correction: this stage is actually just installing the ADOT Python in the `/operator-build` directory. Copying to `/autoinstrumentation` folder is actually done in stage 3.

Dockerfile
```diff
@@ -18,11 +20,42 @@ RUN sed -i "/opentelemetry-exporter-otlp-proto-grpc/d" ./aws-opentelemetry-distr
...
+# Cross-compile based on the target platform.
+RUN if [ $TARGETARCH = "amd64" ]; then export ARCH="x86_64" ; \
+    elif [ $TARGETARCH = "arm64" ]; then export ARCH="aarch64" ; \
+    else false; \
+    fi \
+    && rustup target add ${ARCH}-unknown-linux-musl \
+    && cargo test  --target ${ARCH}-unknown-linux-musl \
+    && cargo install --target ${ARCH}-unknown-linux-musl --path . --root .
+
+# Stage 3: Copy /autoinstrumentation and cp to scratch image
```
srprash, line 54: Correction: This stage is actually building the distribution. It copies the THIRD-PARTY-LICENSES fine, the custom built (from stage 2) `cp` command, and the installed ADOT Python (from stage 1) to their respective destinations within the docker image that we finally distribute.

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+    match operation.copy_type {
+        CopyType::Archive => copy_archive(&operation.source, &operation.destination)?,
+        CopyType::SingleFile => fs::copy(&operation.source, &operation.destination).map(|_| ())?,
+    };
+    Ok(())
+}
+
+// Execute the recursive type of copy operation
+fn copy_archive(source: &Path, dest: &Path) -> io::Result<()> {
+    let mut stack = VecDeque::new();
```
srprash, line 78: https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/tools/cp-utility/src/main.rs#L79-L87

I still don't get how we can have a pre-existing `/otel-auto-instrumentation-python` folder that would cause the code change to fix the problem. The target container shouldn't have this folder to begin with and should only have this when the `adot-autoinstrumentation-python` image is mounted via the `init` container either by the CW Operator or manually.

srprash: Also, if this is the bug fix, we should consider porting the fix back to adot java.
https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/tools/cp-utility/src/main.rs#L79-L87

Can you cut a backlog item for this task?

ignore-words.txt
```diff
@@ -0,0 +1 @@
+crate
```
srprash, line 1: What is this file for and why do we need it?

Dockerfile
harrryr: Fixed

Dockerfile
harrryr: Corrected

ignore-words.txt
harrryr: Removed. Left there previously when importing additional libraries for testing

tools/cp-utility/src/main.rs
harrryr: Yeah I'm not sure why the folder is there either. Running the sample app pod without the operator doesn't have the `/otel-auto-instrumentation-python` folder. Do you think this warrants further investigation?

harrryr: Confirmed that this is expected behavior with Mengyi

ignore-words.txt
harrryr: Update: Sorry they are actually needed because tox spellcheck thinks crate is supposed to be spelled create. Added back

ignore-words.txt
srprash, line 1: Can we instead use the [inline ignore feature](https://github.com/codespell-project/codespell?tab=readme-ov-file#inline-ignore)? Or use [`-L` parameter](https://github.com/codespell-project/codespell?tab=readme-ov-file#ignoring-words) for this word? It's just one word used in one place, so I don't think we need to create a file for that. Plus the ignore words file can't seem to have any comments so we can't give any context here.

srprash (approved):

tox.ini
```diff
@@ -46,7 +46,7 @@ deps =
   codespell
 
 commands =
-  codespell
+  codespell -L crate
```
thpierce, line 49: Can you add to https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/.codespellrc

thpierce (commented):
Please add testing for both AMD64 and ARM64 as we have conditional logic for both.

tox.ini
srprash: Didn't even realize we already had a file for this! :D 
Cool.

harrryr: Done

thpierce (approved):

PR #218, opened 2024-06-21
Add Monkey Patch for Gevent

harrryr:
*Issue #, if available:*
Some customers are using gevent in their Python application with opentelemetry, and this causes the application to crash due to `maximum recursion depth exceeded` error. This is because `gevent` uses a specific version of the `ssl` module and runs `monkey.patch_all()` to patch the ssl module to the required version.

The `monkey` command patches already existing modules to the required version, which may cause problems if there are other applications already using the module before monkey is ran. 

Because Opentelemetry is also using the `ssl` module , it creates a conflict and causes the application to crash. The recommended solution is to run `moneky.patch_all()` in opentelemetry beforehand so that such conflicts do not occur. 

*Description of changes:*
The issue experienced by customers is specifically due to the `ssl` module, therefore, we can run monkey.patch_ssl() command to only monkey the `ssl` module and reduce the scope. The distro will check if the customer application has the `gevent` module installed, and if so, run the `monkey.patch_ssl()`. 

*Test*
Built opentelemetry docker image with the change, modified the operator to install opentelemetry on sample python application with the image. Checked that the sample application has opentelemetry running and doesn't crash when APIs are called. 

Test run: https://github.com/aws-observability/aws-otel-python-instrumentation/actions/runs/9813426624


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
```diff
@@ -20,6 +20,12 @@ def apply_instrumentation_patches() -> None:
 
     Where possible, automated testing should be run to catch upstream changes resulting in broken patches
     """
+    if _is_installed("gevent"):
+        # pylint: disable=import-outside-toplevel
+        # Delay import to only occur if monkey patch is needed (e.g. gevent is used to run application).
+        from gevent import monkey
+
+        monkey.patch_ssl()
```
srprash, line 28: The [gevent.monkey](https://www.gevent.org/api/gevent.monkey.html) doc recommends using `patch_all()`.
>The primary interface to this is the [patch_all()](https://www.gevent.org/api/gevent.monkey.html#gevent.monkey.patch_all) function, which performs all the available patches. It accepts arguments to limit the patching to certain modules, but most programs should use the default values as they receive the most wide-spread testing, and some monkey patches have dependencies on others.

What's the reason we are not using the `patch_all()` here?

*Just thinking out loud here:* 
I don't trust the `gevent.monkey` module mainly because 1) it is a 3rd party code which I don't know much about. 2) It is used in our distro for a very specific user base. So I'm thinking how can we make this safe and possible for users to get out of a possible bad situation caused by this code without having to wait for us to fix and release.

Maybe we can introduce an environment variable (something like `AWS_GEVENT_PATCH_MODULES`??) to control the behavior of this patching? The env variable can accept `all`, `none`, or a comma separated list of modules to patch which can be passed into the [`patch_all()` method](https://www.gevent.org/api/gevent.monkey.html#gevent.monkey.patch_all). By default we patch all if gevent is installed but still give users control to either completely disable this patch or patch select modules.

I know this will make the code a little bit complex and we will have to maintain an an extra environment variable, but it provides improved user experience.

srprash: @thpierce any thoughts on this?

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_instrumentation_patch.py
```diff
@@ -93,6 +96,10 @@ def _test_unpatched_botocore_instrumentation(self):
         self.assertFalse("aws.sqs.queue_url" in attributes)
         self.assertFalse("aws.sqs.queue_name" in attributes)
 
+    def _test_unpatched_gevent_ssl_instrumentation(self):
+        # Ssl
+        self.assertFalse("gevent.ssl" in sys.modules, "Upstream has added the gevent ssl patch")
```
srprash, line 101: How does checking `gevent.ssl` in sys.modules help verify that the `ssl` module was in fact actually patched by the `gevent.monkey` module? Is there a better way to check? Maybe use the [`is_module_patched()`](https://www.gevent.org/api/gevent.monkey.html#gevent.monkey.is_module_patched) method from gevent?

Also, the messaging is incorrect. This patching has nothing to do with the "Upstream" (meaning OTel). It is us that is actually doing the gevent patch.

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_instrumentation_patch.py
```diff
@@ -115,6 +122,10 @@ def _test_patched_botocore_instrumentation(self):
         self.assertTrue("aws.sqs.queue_name" in sqs_attributes)
         self.assertEqual(sqs_attributes["aws.sqs.queue_name"], _QUEUE_NAME)
 
+    def _test_patched_gevent_ssl_instrumentation(self):
+        # Ssl
+        self.assertTrue("gevent.ssl" in sys.modules)
```
srprash, line 127: Same comment as above on improving the check if module was patched or not.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
```diff
@@ -20,6 +20,12 @@ def apply_instrumentation_patches() -> None:
 
     Where possible, automated testing should be run to catch upstream changes resulting in broken patches
     """
+    if _is_installed("gevent"):
```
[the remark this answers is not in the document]
srprash: Ah. `gevent` package versioning is all over the place 🤦 
https://pypi.org/project/gevent/#history

There are major version updates with just bugfixes: https://github.com/gevent/gevent/blob/master/CHANGES.rst

All the more reason to make this piece of code fail safe.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
harrryr: `patch_ssl()` is a sub-command under `patch_all()`. patch_all() also patches some other libraries such as os, sys, time, etc ([link](https://www.gevent.org/_modules/gevent/monkey.html#patch_all)) We're using `patch_ssl()` to minimize the scope of the patch.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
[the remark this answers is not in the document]
harrryr: Added around try-except

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
```diff
@@ -20,6 +25,40 @@ def apply_instrumentation_patches() -> None:
...
+
+                monkey.patch_all()
+            elif gevent_patch_module == "none":
+                pass
+            else:
+                module_list = [module.strip() for module in gevent_patch_module.split(",")]
+
+                # pylint: disable=import-outside-toplevel
+                # Delay import to only occur if monkey patch is needed (e.g. gevent is used to run application).
+                from gevent import monkey
```
srprash, line 44: Try not to repeat statements for better maintainability. You can rearrange the if-else blocks to only import the module once when `gevent_patch_module` is NOT `none`.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
```diff
@@ -20,6 +25,40 @@ def apply_instrumentation_patches() -> None:
...
+                    ssl="ssl" in module_list,
+                    subprocess="subprocess" in module_list,
+                    sys="sys" in module_list,
+                    builtins="builtins" in module_list,
+                    signal="signal" in module_list,
+                    queue="queue" in module_list,
+                    contextvars="contextvars" in module_list,
+                )
+        except Exception as exc:  # pylint: disable=broad-except
+            _logger.debug("Failed to monkey patch gevent, exception: %s", exc)
```
srprash, line 61: I would change it to either info or warning log level. Debug logs aren't printed unless specifically enabled.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
```diff
@@ -65,5 +68,7 @@ def _configure(self, **kwargs):
             OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION, "base2_exponential_bucket_histogram"
         )
 
+        os.environ.setdefault(AWS_GEVENT_PATCH_MODULES, "all")
```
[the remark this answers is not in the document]
harrryr: Done

srprash (approved):

thpierce (approved):

PR #221, opened 2024-06-28
adding mysql-connector driver contract test for some basic operations

BlueLort:
*Issue #, if available:*
N/A

*Description of changes:*
Adding contract tests for mysql-connector driver for MySQL.

Includes `SELECT`, `CREATE DATABASE`, `DROP TABLE` 

```
./scripts/build_and_install_distro.sh && \
./scripts/set-up-contract-tests.sh && \
python3 -m pytest contract-tests/tests/test/amazon/mysql-connector/mysql_connector_test.py

...
...
..
..
======================================================================= warnings summary ========================================================================
<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: Type google._upb._message.MessageMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: Type google._upb._message.ScalarMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================ 4 passed, 2 warnings in 50.84s =================================================================
```
```

### DIFF
--- a/contract-tests/images/applications/mysql-connector/mysql_connector_server.py
+++ b/contract-tests/images/applications/mysql-connector/mysql_connector_server.py
@@ -1,0 +1,90 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+import atexit
+import os
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from threading import Thread
+from typing import Tuple
+
+import mysql.connector as mysql
+
+from typing_extensions import override
+
+_PREPARE_DB: str = "prepare_db"
+_SELECT: str = "select"
+_CREATE_DATABASE: str = "create_database"
+_DROP_TABLE: str = "drop_table"
+_FAULT: str = "fault"
+_PORT: int = 8080
+
+_DB_HOST = os.getenv("DB_HOST")
+_DB_USER = os.getenv("DB_USER")
+_DB_PASS = os.getenv("DB_PASS")
+_DB_NAME = os.getenv("DB_NAME")
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+    @override
+    # pylint: disable=invalid-name
+    def do_GET(self):
+        status_code: int = 200
+        conn = mysql.connect(host=_DB_HOST, user=_DB_USER, password=_DB_PASS, database=_DB_NAME)
+        conn.autocommit = True  # CREATE DATABASE cannot run in a transaction block
+        if self.in_path(_PREPARE_DB):
+            cur = conn.cursor()
+            cur.execute("CREATE TABLE employee (id int, name varchar(255))")
+            cur.execute("INSERT INTO employee (id, name) values (1, 'A')")
+            cur.close()
+            status_code = 200
+        elif self.in_path(_SELECT):
+            cur = conn.cursor()
+            cur.execute("SELECT count(*) FROM employee")
+            result = cur.fetchall()
+            cur.close()
+            status_code = 200 if len(result) == 1 else 500
+        elif self.in_path(_DROP_TABLE):
+            cur = conn.cursor()
+            cur.execute("DROP TABLE IF EXISTS test_table")
+            cur.close()
+            status_code = 200
+        elif self.in_path(_CREATE_DATABASE):
+            cur = conn.cursor()
+            cur.execute("CREATE DATABASE test_database")
+            cur.close()
+            status_code = 200
+        elif self.in_path(_FAULT):
+            cur = conn.cursor()
+            try:
+                cur.execute("SELECT DISTINCT id, name FROM invalid_table")
+            except mysql.ProgrammingError as exception:
+                print("Expected Exception with Invalid SQL occurred:", exception)
+                status_code = 500
+            except Exception as exception:  # pylint: disable=broad-except
+                print("Exception Occurred:", exception)
+            else:
+                status_code = 200
+            finally:
+                cur.close()
+        else:
+            status_code = 404
+        conn.close()
+        self.send_response_only(status_code)
+        self.end_headers()
+
+    def in_path(self, sub_path: str):
+        return sub_path in self.path
+
+
+def main() -> None:
+    server_address: Tuple[str, int] = ("0.0.0.0", _PORT)
+    request_handler_class: type = RequestHandler
+    requests_server: ThreadingHTTPServer = ThreadingHTTPServer(server_address, request_handler_class)
+    atexit.register(requests_server.shutdown)
+    server_thread: Thread = Thread(target=requests_server.serve_forever)
+    server_thread.start()
+    print("Ready")
+    server_thread.join()
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: contract-tests/images/applications/mysql-connector/mysql_connector_server.py [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+    def do_GET(self):
+        status_code: int = 200
+        conn = mysql.connect(host=_DB_HOST, user=_DB_USER, password=_DB_PASS, database=_DB_NAME)
+        conn.autocommit = True  # CREATE DATABASE cannot run in a transaction block
+        if self.in_path(_PREPARE_DB):
+            cur = conn.cursor()
+            cur.execute("CREATE TABLE employee (id int, name varchar(255))")
+            cur.execute("INSERT INTO employee (id, name) values (1, 'A')")
+            cur.close()
+            status_code = 200
```
BlueLort, line 38: Can be discussed looking for a better way than adding an endpoint 
problem is on tests I would have to change a lot in the base tests if all of these operations would be done under SELECT endpoint.

So I call Prepare before select and then clear the mock collector.

contract-tests/tests/test/amazon/base/database_contract_test_base.py
```diff
@@ -55,6 +55,13 @@ def assert_create_database_succeeds(self) -> None:
         self.mock_collector_client.clear_signals()
         self.do_test_requests("create_database", "GET", 200, 0, 0, sql_command="CREATE DATABASE")
 
+    def assert_select_succeeds(self) -> None:
+        self.send_request("GET", "prepare_db")
+        # wait for metrics and traces to be available on the collector before clearing the signals
+        time.sleep(1)
```
BlueLort, line 61: I don't like this is in the code tbh, would like to see other options, I tried to dig deeper into mock collector and how it works but seemed ok to me the way we clear signals.

I couldn't get around it unless I have this sleep till all traces & metrics are received. 

otherwise on assertions later when we query for metrics some metrics are received and tests can get flakey.

Really looking into better solution/discussion on this one.

scripts/set-up-contract-tests.sh
```diff
@@ -18,12 +18,14 @@ rm -rf dist/mock_collector*
 rm -rf dist/contract_tests*
 
 # Install python dependency for contract-test
-pip install pymysql
-pip install cryptography
+python3 -m pip install pytest
```
BlueLort, line 21: adding pytest since i don't have it by default and I think other contributors might face same issue.

the other this is my default python & pip were 2.7 so i'm forcing usage of python3 here.

contract-tests/images/applications/mysql-connector/requirements.txt
vastin, line 1: Please update the versions to align with requirements.txt of psycopg2/pymysql tests. 
```
opentelemetry-distro==0.46b0
opentelemetry-exporter-otlp-proto-grpc==1.25.0

contract-tests/images/applications/mysql-connector/mysql_connector_server.py
vastin: Please explain more about why it does not align with psycopg2/pymysql tests. If it is necessary, please also update the psycopg2/pymysql tests.

contract-tests/tests/test/amazon/base/database_contract_test_base.py
vastin: If it is an enhancement of database test, could it be split into another PR and cover all database test cases?

scripts/set-up-contract-tests.sh
vastin: It makes sense to use `python3 -m pip`.

vastin (changes requested):

contract-tests/tests/test/amazon/base/database_contract_test_base.py
BlueLort: Not really an enhancement, it's an approach to be able to test only Select operation.

For other database drivers, I think can be done as part of another PR when needed to extend, for this one focusing on mysql-connector-driver.

contract-tests/images/applications/mysql-connector/mysql_connector_server.py
BlueLort: I'm creating mysql-connector-driver with additional operation to test which is SELECT.

To be able to test SELECT alone, I have added this prepare db to add data into the DB before firing the SELECT query.

for Other DB drivers I. can once I get agreement on this approach also I think its ok to be part of other PR as well.

contract-tests/images/applications/mysql-connector/requirements.txt
BlueLort: cool, sorry this PR has been there for a week, will update it

contract-tests/images/applications/mysql-connector/mysql_connector_server.py
BlueLort: Addressed all DB Drivers to have the same test.

contract-tests/tests/test/amazon/base/database_contract_test_base.py
vastin: Not sure if time.sleep(1) is long enough or overkill. Could the table preparation be split into a new test case instead of mixing with select test?

BlueLort: I tried with 0.5 and it was sometimes flakey not that common, so i doubled the number.

I was thinking of multiple approaches to resolve this, also i was trying to understand in-depth how the mock collector works and it seemed normal to me that it flushes once we ask it to clear signals.

Separating this into another testcase will not still prepare the db with data needed for select operation test as every test spin off a new test container.

additionally im not testing the prepartion im actually testing select.

Having the preparation done anyway on the server will result into adding more traces into the mock collector and wont give us the granualty to check for SELECT specifically.


Also i was thinking about changing the Query itself, but I think I should follow like an average query a consumer of agent would use.

I'm open for any idea to make it better i don't like the sleep also here. do you know more about mock collectors and why clear signals are not enough in this case? 👀

vastin: If only testing select, could the db table is prepared in mysql_connector_server app before accepting incoming http requests?

vastin: For mock collector clear signals issue, probably @thpierce has more insight.

BlueLort: test container gets spinned off for the server while testing, which will show this prepartion flow in the traces, same as adding preperation on the SELECT endpoint.

checking with @thpierce for insights :)

thpierce: Suspect the correct approach here is to create and populate the table when the container starts up, similar to how we are doing things in the [botocore tests](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/contract-tests/images/applications/botocore/botocore_server.py#L215)

If the problem is that `prepare_db` is returning and telemetry isn't being created for up to 1 second, I would suspect that there is something wrong with how we have implemented prepare_db. Perhaps you are invoking something asynchronous, and telemetry is not being created until the mySQL request finishes, which is well after the API returns. If this is the case, we should avoid such async behaviour where possible.

vastin (approved):
LGTM

PR #227, opened 2024-07-11
Add bedrock contract tests.

zzhlogin:
The PR is a follow up on bedrock service support PR:
https://github.com/aws-observability/aws-otel-python-instrumentation/pull/209

We add contract tests for following Bedrock services that covers all resource attributes we newly support:
1. Bedrock API: `GetGuardrail` 
2. BedrockAgent APIs: `GetAgent`, `GetDataSource`, `GetKnowledgeBase`
3. BedrockRuntime API: `InvokeModel`
4. BedrockAgentRuntime API: `InvokeAgent`

Upgrade `botocore` and `boto3` to the latest version `1.34.143` so that to support Bedrock services API calls.
Upgrade `localstack/localstack` image to the latest version `3.5.0` so resolve the SQS API call issue using `localstack/localstack:2.0.1` with new version of `boto3`: https://github.com/localstack/localstack/issues/9610

**Contract test limitation:**
The contract tests in current repo is using [LocalStackContainer](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/912dd93ff19b7a4594bb9ed1a7d8cde4907a735d/contract-tests/tests/test/amazon/botocore/botocore_test.py#L68) to serve AWS SDK service calls. But it doesn’t has bedrock related service support (This is [the full service list](https://docs.localstack.cloud/references/coverage/) it support.). In this case, no matter which bedrock API we call in contract test, the response will always be 4XX.

As a workaround, we inject `inject_200_success` and `inject_500_error` directly into the API call to make sure we receive http response with expected status code and attributes. 

**`_assert_semantic_conventions_span_attributes` function change:**
In [_assert_semantic_conventions_span_attributes function](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/1753bbf2c3cd41778abc358a7f1e3199e48c7fa9/contract-tests/tests/test/amazon/botocore/botocore_test.py#L448) it is checking the input `service` equals to `rpc.service`, however, we pass the input service with ["remote_service"](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/1753bbf2c3cd41778abc358a7f1e3199e48c7fa9/contract-tests/tests/test/amazon/botocore/botocore_test.py#L430), where there is mismatch for example for Bedrock Agent Runtime: we have
`rpc_service="Bedrock Agent Runtime", remote_service="AWS::Bedrock". `
Thus we change to use `rpc_service` if it is provided by:
```
kwargs.get("rpc_service") if "rpc_service" in kwargs else kwargs.get("remote_service").split("::")[-1],
```

### DIFF
--- a/contract-tests/tests/test/amazon/botocore/botocore_test.py
+++ b/contract-tests/tests/test/amazon/botocore/botocore_test.py
@@ -31,8 +31,8 @@ _AWS_SQS_QUEUE_NAME: str = "aws.sqs.queue.name"
 _AWS_KINESIS_STREAM_NAME: str = "aws.kinesis.stream.name"
 _AWS_BEDROCK_AGENT_ID: str = "aws.bedrock.agent.id"
 _AWS_BEDROCK_GUARDRAIL_ID: str = "aws.bedrock.guardrail.id"
-_AWS_BEDROCK_KNOWLEDGEBASE_ID: str = "aws.bedrock.knowledge_base.id"
-_AWS_BEDROCK_DATASOURCE_ID: str = "aws.bedrock.data_source.id"
+_AWS_BEDROCK_KNOWLEDGE_BASE_ID: str = "aws.bedrock.knowledge_base.id"
+_AWS_BEDROCK_DATA_SOURCE_ID: str = "aws.bedrock.data_source.id"
 _GEN_AI_REQUEST_MODEL: str = "gen_ai.request.model"
 
 
@@ -377,7 +377,7 @@ class BotocoreTest(ContractTestBase):
             span_name="Kinesis.PutRecord",
         )
 
-    def test_bedrock_invoke_model(self):
+    def test_bedrock_runtime_invoke_model(self):
         self.do_test_requests(
             "bedrock/invokemodel/invoke-model",
             "GET",
@@ -393,24 +393,6 @@ class BotocoreTest(ContractTestBase):
                 _GEN_AI_REQUEST_MODEL: "amazon.titan-text-premier-v1:0",
             },
             span_name="Bedrock Runtime.InvokeModel",
-        )
-
-    def test_bedrock_get_agent(self):
-        self.do_test_requests(
-            "bedrock/getagent/get-agent",
-            "GET",
-            200,
-            0,
-            0,
-            rpc_service="Bedrock Agent",
-            remote_service="AWS::Bedrock",
-            remote_operation="GetAgent",
-            remote_resource_type="AWS::Bedrock::Agent",
-            remote_resource_identifier="TESTAGENTID",
-            request_specific_attributes={
-                _AWS_BEDROCK_AGENT_ID: "TESTAGENTID",
-            },
-            span_name="Bedrock Agent.GetAgent",
         )
 
     def test_bedrock_get_guardrail(self):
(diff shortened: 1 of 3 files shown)


mxiamxia (commented):
in general, the pr looks good to me. the tests covers the new gen ai attributes will be released.

mxiamxia (approved):

contract-tests/tests/test/amazon/botocore/botocore_test.py
```diff
@@ -372,6 +377,132 @@ def test_kinesis_fault(self):
...
         )
 
+    def test_bedrock_runtime_invoke_model(self):
+        self.do_test_requests(
+            "bedrock/invokemodel/invoke-model",
+            "GET",
+            200,
+            0,
+            0,
+            rpc_service="Bedrock Runtime",
```
mxiamxia, line 387: Q: why this is required for Bedrock/runtime service but not for the other existing supported AWS resources? such as Kinesis, etc.

zzhlogin: As in PR description:

> **`_assert_semantic_conventions_span_attributes` function change:**
> In [_assert_semantic_conventions_span_attributes function](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/1753bbf2c3cd41778abc358a7f1e3199e48c7fa9/contract-tests/tests/test/amazon/botocore/botocore_test.py#L448) it is checking the input `service` equals to `rpc.service`, however, we pass the input service with ["remote_service"](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/1753bbf2c3cd41778abc358a7f1e3199e48c7fa9/contract-tests/tests/test/amazon/botocore/botocore_test.py#L430), where there is mismatch for example for Bedrock Agent Runtime: we have
> `rpc_service="Bedrock Agent Runtime", remote_service="AWS::Bedrock". `
> Thus we change to use `rpc_service` if it is provided by:
> ```
> kwargs.get("rpc_service") if "rpc_service" in kwargs else kwargs.get("remote_service").split("::")[-1],
> ```

Whereas, for the other services, for example Kinesis, the `remote_service` is the same as `rpc_service`

mxiamxia: The reason it is not needed for S3 and other supported services because the `rpc_service` value always matches the second part(after `::`) of value in `remote_service`?

contract-tests/tests/test/amazon/botocore/botocore_test.py
```diff
@@ -372,6 +377,132 @@ def test_kinesis_fault(self):
...
+            0,
+            rpc_service="Bedrock Agent",
+            remote_service="AWS::Bedrock",
+            remote_operation="GetDataSource",
+            remote_resource_type="AWS::Bedrock::DataSource",
+            remote_resource_identifier="DATASURCID",
+            request_specific_attributes={
+                _AWS_BEDROCK_DATA_SOURCE_ID: "DATASURCID",
+            },
+            span_name="Bedrock Agent.GetDataSource",
```
thpierce, line 503: Not blocking: Slightly cleaner to do the following
```suggestion
            span_name=f'{rpc_service}.{remote_operation}'
```

contract-tests/tests/test/amazon/botocore/botocore_test.py
zzhlogin: Yes.

PR #243, opened 2024-08-28
Opt-in config and logic for deferral of instrumentation to only WSGI worker processes

srprash:
### Issue:
OTel Python [has issues](https://github.com/open-telemetry/opentelemetry-python/issues/2767) where the SDK is unable to report metrics for applications using a fork process model WSGI server.
This affects ADOT when it tries to generate the OTel or Application Signals metrics. 

A solution to this is to [re-initialize the SDK in the worker processes after the process forking as happened](https://opentelemetry-python.readthedocs.io/en/latest/examples/fork-process-model/README.html). A small caveat is that if the SDK has been initialized in the master process, the worker process SDK won't work because Tracer/Meter providers can be set globally only once. So to circumvent this, we need to skip initializing the SDK in the master process and only do so in the worker processes.

### Description of changes:
- Introducing an opt-in configuration environment variable `OTEL_AWS_PYTHON_DEFER_TO_WORKERS_ENABLED` to enable if they are using a WSGI (or a fork process model) server and want the ADOT SDK to defer auto-instrumentation to worker processes.
- Whenever the ADOT SDK auto-instrumentation is loaded (either via the `sitecustomize.py` file or the `opentelemetry-instrument` command), the SDK will check if the above configuration is enabled and if the current process is the master process, and will skip the instrumentation. 
- The way we determine if the current process is master or worker is by using an internal marker environment variable `IS_WSGI_MASTER_PROCESS_ALREADY_SEEN`. The first time the ADOT SDK sees a python process, this env var is not set and it will know this should be a WSGI master process. We then set the env var and when a new worker process forks, the master environment is copied to it (and so the env var). So when the ADOT SDK checks this env var again (in worker) it finds that the env var was already set to `true` in the master.

### Testing:
- Unit tests covering the functionalities bases on different configurations of the `OTEL_AWS_PYTHON_DEFER_TO_WORKERS_ENABLED` and `IS_WSGI_MASTER_PROCESS_ALREADY_SEEN` variables.
- Manual test using a sample application. Since this is an opt-in configuration (a 2-way door), testing manually gives us a fair bit of confidence.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
@@ -65,6 +65,7 @@ METRIC_EXPORT_INTERVAL_CONFIG = "OTEL_METRIC_EXPORT_INTERVAL"
 DEFAULT_METRIC_EXPORT_INTERVAL = 60000.0
 AWS_LAMBDA_FUNCTION_NAME_CONFIG = "AWS_LAMBDA_FUNCTION_NAME"
 AWS_XRAY_DAEMON_ADDRESS_CONFIG = "AWS_XRAY_DAEMON_ADDRESS"
+OTEL_AWS_PYTHON_DEFER_TO_WORKERS_ENABLED_CONFIG = "OTEL_AWS_PYTHON_DEFER_TO_WORKERS_ENABLED"
 
 _logger: Logger = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -85,6 +86,8 @@ class AwsOpenTelemetryConfigurator(_OTelSDKConfigurator):
     # pylint: disable=no-self-use
     @override
     def _configure(self, **kwargs):
+        if _is_defer_to_workers_enabled() and _is_wsgi_master_process():
+            return
```
mxiamxia, line 90: Can we add a `info` level log message when skipping?

mxiamxia: And `pid` in the log message as well?

srprash: Good idea. will add that rn.

mxiamxia (approved):
thank you!

thpierce (approved):

PR #247, opened 2024-08-30
Export unsampled span in AWS Lambda environment

wangzlei:
*Description of changes:*

When AppSignals is enabled in AWS Lambda, we export both sampled and unsampled spans to X-Ray to generate Application Signals metrics.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_batch_unsampled_span_processor.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_batch_unsampled_span_processor.py
@@ -1,0 +1,37 @@
+import logging
+from typing import Optional
+
+from amazon.opentelemetry.distro._aws_attribute_keys import AWS_TRACE_FLAG_UNSAMPLED
+from opentelemetry.context import Context
+from opentelemetry.sdk.trace import ReadableSpan, Span
+from opentelemetry.sdk.trace.export import BatchSpanProcessor as BaseBatchSpanProcessor
+
+logger = logging.getLogger(__name__)
+
+SPANUNSAMPLED_FLAG = "OTEL_AWS_APP_SIGNALS_ENABLED"
+
+
+class BatchUnsampledSpanProcessor(BaseBatchSpanProcessor):
+
+    def on_start(self, span: Span, parent_context: Optional[Context] = None) -> None:
+        if not span.context.trace_flags.sampled:
+            span.set_attribute(AWS_TRACE_FLAG_UNSAMPLED, "True")
+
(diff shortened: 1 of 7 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_batch_unsampled_span_processor.py
```diff
@@ -0,0 +1,42 @@
...
+
+SPANUNSAMPLED_FLAG = "OTEL_AWS_APP_SIGNALS_ENABLED"
+
+
+class BatchUnsampledSpanProcessor(BaseBatchSpanProcessor):
+
+    # pylint: disable=no-self-use
+    def on_start(self, span: Span, parent_context: Optional[Context] = None) -> None:
+        if not span.context.trace_flags.sampled:
+            span.set_attribute(AWS_TRACE_FLAG_UNSAMPLED, "True")
```
mxiamxia, line 21: should we use lower case `True` as attribute value which will be visible in Spans? It looks all other boolean attribute s are using lower case convention. Eg. `"aws.is.local.root":true,`

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_batch_unsampled_span_processor.py
```diff
@@ -0,0 +1,42 @@
...
+        if self.done:
+            logger.warning("Already shutdown, dropping span.")
+            return
+
+        if len(self.queue) == self.max_queue_size:
+            # pylint: disable=access-member-before-definition
+            if not self._spans_dropped:
+                logger.warning("Queue is full, likely spans will be dropped.")
+                # pylint: disable=attribute-defined-outside-init
+                self._spans_dropped = True
```
mxiamxia, line 36: Q: what will we do after marking `self._spans_dropped = True` on the dropped span?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -154,6 +155,7 @@ def _init_tracing(
         span_exporter = _customize_exporter(span_exporter, resource)
         trace_provider.add_span_processor(BatchSpanProcessor(span_exporter))
 
+    _export_unsampled_span_for_lambda(trace_provider, resource)
```
mxiamxia, line 158: nit: can we place this logic inside `_customize_span_processors` function where we set the custom span processors for trace provider. It seems will be organized. A picky comment :-0

mxiamxia (approved):

PR #261, opened 2024-09-20
feat: Support a list of new AWS Resources for Python SDK

yiyuan-he:
### *Description of changes:*
Adding support for SNS AWS resource for Python SDK with auto-instrumentation.

Specifically, this change populates `aws.sns.topic.arn` in the Span by extracting the `TopicArn` field from `SnsRequest`.
- CFNIdentifier: `"primaryIdentifier": [ "/properties/TopicArn" ]`

### *Testing Plan:*
Set up a simple client-server locally and verified that the new data is added to the span.

Screenshot 2024-09-20 at 12 16 59 PM

Ran Code Style Checks:
```
codespell . --write-changes
python scripts/eachdist.py lint
```

Added unit tests for:
- AWS Metric Attribute Generator
- Instrumentation Patches

Screenshot 2024-09-20 at 12 23 46 PM

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #264, opened 2024-09-24
feat: Add CFN Primary ID for Existing Supported AWS Resources

yiyuan-he:
### *Description of changes:*

These changes support AWS Resources with different CFN Identifier formats. This new attribute will have the key `aws.remote.resource.cfn.primary.identifier`.

More context can be found in the comments of this closed PR: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/261

In most cases, this new attribute should have the same value as the existing attribute `aws.remote.resource.identifier`.

However, there are some edge cases we must handle as seen in this PR:
- `AWS::SQS::Queue`
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html
  - The CFN Id should be the URL of the Queue.
- `AWS::Bedrock::DataSource`
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-bedrock-datasource.html
  - The CFN Id should be a compound value where the knowledge base Id and the data source Id are separated by `|`

Also updated Bedrock Agent request parameters to also include Knowledge Base ID: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agent/client/get_data_source.html

### *Test Plan:*
Set up a client-server with auto-instrumentation to verify that the correct span data is being generated.

bedrock_cfn_primary_id_verification
sqs_cfn_primary_id_verification

Unit tests for instrumentation
metric_generator_unit_test_verification
instrumentation_patch_unit_test_verification


mxiamxia (approved):
LGTM! ty!

PR #283, opened 2024-10-18
feat: update contract tests for cfn primary id [NOT READY]

yiyuan-he:
*Description of changes:*
Updating our contract tests to verify the new cloudformation primary id attribute.

### DIFF
--- a/contract-tests/tests/test/amazon/botocore/botocore_test.py
+++ b/contract-tests/tests/test/amazon/botocore/botocore_test.py
@@ -10,6 +10,7 @@ from typing_extensions import override
 
 from amazon.base.contract_test_base import NETWORK_NAME, ContractTestBase
 from amazon.utils.application_signals_constants import (
+    AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER,
     AWS_LOCAL_OPERATION,
     AWS_LOCAL_SERVICE,
     AWS_REMOTE_OPERATION,
@@ -99,6 +100,7 @@ class BotocoreTest(ContractTestBase):
             remote_operation="CreateBucket",
             remote_resource_type="AWS::S3::Bucket",
             remote_resource_identifier="test-bucket-name",
+            cloudformation_primary_identifier="test-bucket-name",
             request_specific_attributes={
                 SpanAttributes.AWS_S3_BUCKET: "test-bucket-name",
             },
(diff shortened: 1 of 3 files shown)


contract-tests/tests/test/amazon/botocore/botocore_test.py
```diff
@@ -497,12 +529,153 @@ def test_bedrock_agent_get_data_source(self):
...
+            cloudformation_primary_identifier=(
+                "arn:aws:states:us-west-2:000000000000:stateMachine:invalid-state-machine"
+            ),
+            request_specific_attributes={
+                _AWS_STATE_MACHINE_ARN: "arn:aws:states:us-west-2:000000000000:stateMachine:invalid-state-machine",
+            },
+            span_name="SFN.ListStateMachineVersions",
+        )
+
+    # TODO: Add contract test for lambda event source mapping resource
```
yiyuan-he, line 677: I will implement this in a separate PR.

**Context:**
When setting up the contract tests, we can create an `EventSourceMapping` resource which will have a randomly generated UUID.

In this test case, we will need to make a `GET` call on this resource and [pass in the randomly generated UUID as an argument.](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/get_event_source_mapping.html)

We need to figure out a way to create some consistency between the randomly generated UUID or else our `GET` call will throw an error.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_botocore_patches.py
```diff
@@ -178,7 +178,16 @@ def patch_extract_attributes(self, attributes: _AttributeMapT):
         if queue_url:
             attributes[AWS_SQS_QUEUE_URL] = queue_url
 
+    old_on_success = _SqsExtension.on_success
+
+    def patch_on_success(self, span: Span, result: _BotoResultT):
```
mxiamxia, line 183: is this a case we missed in the previous implementation?

mxiamxia (approved):

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_botocore_patches.py
yiyuan-he: Yes this affects APIs like [SqsCreateQueue](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/create_queue.html), where the `QueueUrl` is only available in the response body.

PR #289, opened 2024-11-01
Get Sha256 Checksum for Release Artifacts

harrryr:
*Issue #, if available:*
We want to provide the checksum of release artifact to customers.

*Description of changes:*
When new artifact is released through `release_build.yml`, get the sha256 checksum of the artifact then attach it to the release note as a `.sha256` file

Example: [Release Note](https://github.com/harrryr/aws-otel-python-instrumentation/releases/tag/v)

*Testing*
Workflow run: https://github.com/harrryr/aws-otel-python-instrumentation/actions/runs/12384286607
Verified that the .sha256 has the correct content and running `shasum -a 256 -c <artifact_name>.sha256` returns the output `<artifact_name>: OK`

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -120,6 +120,14 @@ jobs:
           tags: |
             ${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ github.event.inputs.version }}
 
+      - name: Get SHA256 checksum of wheel file
+        id: get_sha256
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          filename="aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl"
+          shasum -a 256 dist/$filename | awk '{print $1}' > $filename.sha256
+
       # Publish to GitHub releases
       - name: Create GH release
         id: create_release
@@ -130,4 +138,5 @@ jobs:
              --title "Release v${{ github.event.inputs.version }}" \
              --draft \
              "v${{ github.event.inputs.version }}" \
-             dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl
+             dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl \
+             aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl.sha256
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -14,6 +14,7 @@ env:
   RELEASE_PRIVATE_REPOSITORY: 020628701572.dkr.ecr.us-west-2.amazonaws.com/adot-autoinstrumentation-python
   RELEASE_PRIVATE_REGISTRY: 020628701572.dkr.ecr.us-west-2.amazonaws.com
   PACKAGE_NAME: aws-opentelemetry-distro
+  ARTIFACT_NAME: aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl
 
 permissions:
   id-token: write
@@ -88,7 +89,7 @@ jobs:
           TWINE_USERNAME: '__token__'
           TWINE_PASSWORD: ${{ env.TEST_PYPI_TOKEN_API_TOKEN }}
         run: |
-          twine upload --repository testpypi --skip-existing --verbose dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl
+          twine upload --repository testpypi --skip-existing --verbose dist/${{ env.ARTIFACT_NAME }}
 
       # Publish to prod PyPI
       - name: Publish to PyPI
@@ -96,7 +97,7 @@ jobs:
           TWINE_USERNAME: '__token__'
           TWINE_PASSWORD: ${{ env.PROD_PYPI_TOKEN_API_TOKEN }}
         run: |
-          twine upload --skip-existing --verbose dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl
+          twine upload --skip-existing --verbose dist/${{ env.ARTIFACT_NAME }}
 
       # Publish to public ECR
       - name: Build and push public ECR image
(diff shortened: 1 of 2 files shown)


majanjua-amzn (dismissed):

srprash:
A few questions:
- Where and how will generated checksum be used? 
- There's already a [`version.py`](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/version.py#L4) where the current released version is tracked (with a `.dev0` suffix. Why can't we reuse that instead of creating a new file?

Happy to sync up on this.

.github/workflows/post_release_version_bump.yml
```diff
@@ -102,6 +101,25 @@ jobs:
...
+  
+      - name: Append latest release checksum to release-build-metadata.json
+        run: |
+          ARTIFACT_NAME="aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl"
+          curl -L -o $ARTIFACT_NAME https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v${{ github.event.inputs.version }}/$ARTIFACT_NAME
+
+          CHECKSUM=$(shasum -a 256 $ARTIFACT_NAME | awk '{ print $1 }')
+
+          FILE="release-build-metadata.json"
+          UPDATED_JSON=$(jq --arg version "${{ env.VERSION }}" \
```
srprash, line 113: use either `github.event.inputs.version` or `env.VERSION` throughout. Mixing them is confusing and prone to bugs.

.github/workflows/post_release_version_bump.yml
```diff
@@ -102,6 +101,25 @@ jobs:
...
+                            --arg name "$ARTIFACT_NAME" \
+                            --arg checksum "$CHECKSUM" \
+                            '.release += [{"version": $version, "checksum": [{"name": $name, "checksum": $checksum}]}]' "$FILE")
```
srprash, line 116: Do we really need to append new version entry to the metadata file or can we just overwrite with the latest one? Do we have a use case of maintaining the metadata for older versions? 
With time, the file will grow big and more expensive to parse.

.github/workflows/post_release_version_bump.yml
```diff
@@ -102,6 +101,25 @@ jobs:
...
           git add .github/workflows/daily_scan.yml
+  
+      - name: Append latest release checksum to release-build-metadata.json
+        run: |
+          ARTIFACT_NAME="aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl"
+          curl -L -o $ARTIFACT_NAME https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v${{ github.event.inputs.version }}/$ARTIFACT_NAME
+
+          CHECKSUM=$(shasum -a 256 $ARTIFACT_NAME | awk '{ print $1 }')
+
+          FILE="release-build-metadata.json"
```
srprash, line 112: `release-build-metadata` doesn't explain the purpose of the file. Maybe `release-artifacts-metadata.json`?

.github/workflows/post_release_version_bump.yml
srprash: Thinking a little more on this. If we do need to maintain the checksum for each released version, would storing the checksum file alongside the artifact within the GH release (like [this one](https://github.com/aws-observability/aws-otel-python-instrumentation/releases/tag/v0.7.0)) itself work?
This way, any user can use this checksum to validate the integrity of the artifact.

.github/workflows/post_release_version_bump.yml
harrryr: Ack

.github/workflows/post_release_version_bump.yml
harrryr: > Do we really need to append new version entry to the metadata file or can we just overwrite with the latest one? Do we have a use case of maintaining the metadata for older versions? With time, the file will grow big and more expensive to parse.

That's a good point. I decided to maintain the metadata for the older versions since customers may still be using them and the older versions are still available even if we release new ones so thought it's better to have a record of them. I'm neutral if we want to just overwrite with the new ones though



> Thinking a little more on this. If we do need to maintain the checksum for each released version, would storing the checksum file alongside the artifact within the GH release (like [this one](https://github.com/aws-observability/aws-otel-python-instrumentation/releases/tag/v0.7.0)) itself work? This way, any user can use this checksum to validate the integrity of the artifact.

That was a consideration we made and we decided to go with the `.json` file because it's safer to retrieve programmatically. Also for cases like Dotnet where there is multiple artifacts, didn't want to overfill the release page with just checksum values

.github/workflows/post_release_version_bump.yml
harrryr: Ack

srprash (approved):

majanjua-amzn (approved):

.github/workflows/release_build.yml
```diff
@@ -120,6 +120,14 @@ jobs:
           tags: |
             ${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ github.event.inputs.version }}
 
+      - name: Get SHA256 checksum of wheel file
+        id: get_sha256
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          filename="aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl"
```
majanjua-amzn, line 128: nit: `filename` is used in more than one place, could be worth setting it to an environment variable for reuse to avoid any mistakes in the future

majanjua-amzn (approved):

srprash (commented):
Non-blocking: I would also like to see steps somewhere on how can the checksum file be used. Include a simple command example in the readme?

.github/workflows/release_build.yml
```diff
@@ -130,4 +138,5 @@ jobs:
              --title "Release v${{ github.event.inputs.version }}" \
              --draft \
              "v${{ github.event.inputs.version }}" \
-             dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl
+             dist/aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl \
+             aws_opentelemetry_distro-${{ github.event.inputs.version }}-py3-none-any.whl.sha256
```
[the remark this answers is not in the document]
harrryr: Ack

.github/workflows/release_build.yml
```diff
@@ -120,6 +120,14 @@ jobs:
           tags: |
             ${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ github.event.inputs.version }}
 
+      - name: Get SHA256 checksum of wheel file
+        id: get_sha256
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
harrryr: Removed

srprash (approved):

PR #290, opened 2024-11-08
Gen-AI python implementation

Jeel-mehta:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
@@ -10,10 +10,6 @@ from typing import Any, Dict, Optional
 
 from botocore.response import StreamingBody
 
-_logger = logging.getLogger(__name__)
-# Set logger level to DEBUG
-_logger.setLevel(logging.DEBUG)
-
 from amazon.opentelemetry.distro._aws_attribute_keys import (
     AWS_BEDROCK_AGENT_ID,
     AWS_BEDROCK_DATA_SOURCE_ID,
@@ -46,6 +42,10 @@ _GUARDRAIL_ID: str = "guardrailId"
 _GUARDRAIL_ARN: str = "guardrailArn"
 _MODEL_ID: str = "modelId"
 _AWS_BEDROCK_SYSTEM: str = "aws_bedrock"
+
+_logger = logging.getLogger(__name__)
+# Set logger level to DEBUG
+_logger.setLevel(logging.DEBUG)
 
 
 class _BedrockAgentOperation(abc.ABC):
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
@@ -4,10 +4,15 @@ import abc
 import inspect
 import io
 import json
+import logging
 import math
 from typing import Any, Dict, Optional
 
 from botocore.response import StreamingBody
+
+_logger = logging.getLogger(__name__)
+# Set logger level to DEBUG
+_logger.setLevel(logging.DEBUG)
 
 from amazon.opentelemetry.distro._aws_attribute_keys import (
     AWS_BEDROCK_AGENT_ID,
@@ -275,7 +280,7 @@ class _BedrockRuntimeExtension(_AwsSdkExtension):
                         self._extract_mistral_attributes(attributes, request_body)
 
                 except json.JSONDecodeError:
-                    print("Error: Unable to parse the body as JSON")
+                    _logger.debug("Error: Unable to parse the body as JSON")
 
     def _extract_titan_attributes(self, attributes, request_body):
         config = request_body.get("textGenerationConfig", {})
@@ -352,9 +357,9 @@ class _BedrockRuntimeExtension(_AwsSdkExtension):
                 result["body"] = StreamingBody(new_stream, len(body_content))
 
             except json.JSONDecodeError:
-                print("Error: Unable to parse the response body as JSON")
+                _logger.debug("Error: Unable to parse the response body as JSON")
             except Exception as e:  # pylint: disable=broad-exception-caught, invalid-name
-                print(f"Error processing response: {str(e)}")
+                _logger.debug(f"Error processing response: {str(e)}")
             finally:
                 if original_body is not None:
                     original_body.close()
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
@@ -2,7 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 import abc
 import inspect
-from typing import Dict, Optional
+from typing import Any, Dict, Optional
+import json
+from botocore.response import StreamingBody
+import math
 
 from amazon.opentelemetry.distro._aws_attribute_keys import (
     AWS_BEDROCK_AGENT_ID,
@@ -11,7 +14,7 @@ from amazon.opentelemetry.distro._aws_attribute_keys import (
     AWS_BEDROCK_GUARDRAIL_ID,
     AWS_BEDROCK_KNOWLEDGE_BASE_ID,
 )
-from amazon.opentelemetry.distro._aws_span_processing_util import GEN_AI_REQUEST_MODEL, GEN_AI_SYSTEM
+from amazon.opentelemetry.distro._aws_span_processing_util import GEN_AI_REQUEST_MODEL, GEN_AI_SYSTEM, GEN_AI_REQUEST_MAX_TOKENS, GEN_AI_REQUEST_TEMPERATURE, GEN_AI_REQUEST_TOP_P, GEN_AI_RESPONSE_FINISH_REASONS, GEN_AI_USAGE_INPUT_TOKENS, GEN_AI_USAGE_OUTPUT_TOKENS
 from opentelemetry.instrumentation.botocore.extensions.types import (
     _AttributeMapT,
     _AwsSdkCallContext,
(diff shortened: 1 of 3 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -238,5 +241,164 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+                span.set_attribute(GEN_AI_RESPONSE_FINISH_REASONS, [choices['finish_reason']])
+
+    def _handle_meta_llama_response(self, span: Span, response_body: Dict[str, Any]):
+        #print("This is the response body :", response_body)
+        if 'prompt_token_count' in response_body:
+            span.set_attribute(GEN_AI_USAGE_INPUT_TOKENS, response_body['prompt_token_count'])
+        if 'generation_token_count' in response_body:
+            span.set_attribute(GEN_AI_USAGE_OUTPUT_TOKENS, response_body['generation_token_count'])
+        if 'stop_reason' in response_body:
+            span.set_attribute(GEN_AI_RESPONSE_FINISH_REASONS, response_body['stop_reason'])
```
liustve, line 396: super small but the GEN_AI_RESPONSE_FINISH_REASONS for meta llama should be wrapped inside an array

yiyuan-he:
Workflow builds are failing. Could you take a [look](
https://github.com/aws-observability/aws-otel-python-instrumentation/pull/290/checks)?

They are probably just linting failures which you can resolve by following the instructions under *Code Style Check* [here](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/CONTRIBUTING.md)

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
Jeel-mehta: Hey Steve,
Thanks for catching that, I have updated the code

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -238,5 +251,163 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+        self._set_if_not_none(attributes, GEN_AI_REQUEST_TEMPERATURE, request_body.get("temperature"))
+        self._set_if_not_none(attributes, GEN_AI_REQUEST_TOP_P, request_body.get("top_p"))
+
+    @staticmethod
+    def _set_if_not_none(attributes, key, value):
+        if value is not None:
+            attributes[key] = value
+
+    def on_success(self, span: Span, result: Dict[str, Any]):
+        super().on_success(span, result)
```
yiyuan-he, line 325: I don't think this call to `super` constructor is necessary here. Can you remove this and test to confirm?

Jeel-mehta: Okay, I will do that and tru to run it locally first

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -240,3 +253,155 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+        model_id = self._call_context.params.get(_MODEL_ID)
+        if not model_id:
+            return
+
+        if "body" in result and isinstance(result["body"], StreamingBody):
+            try:
+                # Read the entire content of the StreamingBody
+                body_content = result["body"].read()
+                # Decode the bytes to string and parse as JSON
+                response_body = json.loads(body_content.decode("utf-8"))
```
yiyuan-he, line 331: `StreamingBody` is one time use. We need to replenish it for downstream use after we have used it in our instrumentation logic.

For example, if we want to print the response of the LLM in our application, it will not be available anymore with the current logic.

To fix this we have to add additional logic to copy and replace the stream for downstream use. I've written some sample code you can adapt here:
```
    def on_success(self, span: Span, result: Dict[str, Any]):
        model_id = self._call_context.params.get(_MODEL_ID)

        if not model_id:
            return

        if 'body' in result and isinstance(result['body'], StreamingBody):
            try:
                original_body = result['body']
                body_content = original_body.read()

                # Use one stream for telemetry
                stream = io.BytesIO(body_content)
                telemetry_content = stream.read()
                response_body = json.loads(telemetry_content.decode('utf-8'))
                if "amazon.titan" in model_id:
                    self._handle_amazon_titan_response(span, response_body)

                # Replenish the stream for downstream use in application
                new_stream = io.BytesIO(body_content)
                result['body'] = StreamingBody(new_stream, len(body_content))
            except json.JSONDecodeError:
                print("Error: Unable to parse the response body as JSON")
            except Exception as e:
                print(f"Error processing the response: {str(e)}")
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -240,3 +253,161 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+            finally:
+                # Make sure to close the stream
+                result["body"].close()
+
+    # pylint: disable=no-self-use
+    def _handle_amazon_titan_response(self, span: Span, response_body: Dict[str, Any]):
+        if "inputTextTokenCount" in response_body:
+            span.set_attribute(GEN_AI_USAGE_INPUT_TOKENS, response_body["inputTextTokenCount"])
+
+            result = response_body["results"][0]
```
yiyuan-he, line 359: Trying to access element with `[0]` will throw an error if the `results` array doesn't exist. I think we need another conditional check here.

For example:
```
if 'results' in response_body:
   result = response_body["results"][0]
```
or something more concise like:
```
result = response_body.get('results', [{}])
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
Jeel-mehta: I have tried and tested implementing this on my local environment and have raised a PR for the same.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
Jeel-mehta: I just saw this, I will fix this and make another commit

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
Jeel-mehta: I have tested this and we don't really require this

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -342,21 +347,24 @@ def on_success(self, span: Span, result: Dict[str, Any]):
...
     def _handle_amazon_titan_response(self, span: Span, response_body: Dict[str, Any]):
         if "inputTextTokenCount" in response_body:
             span.set_attribute(GEN_AI_USAGE_INPUT_TOKENS, response_body["inputTextTokenCount"])
-
-            result = response_body["results"][0]
+            if "results" in response_body:
+                result = response_body["results"][0]
             if "tokenCount" in result:
                 span.set_attribute(GEN_AI_USAGE_OUTPUT_TOKENS, result["tokenCount"])
             if "completionReason" in result:
```
yiyuan-he, line 370: I think this logic will still throw an error. 

For example, if `results` does not exist in the `response_body`, the checks will still continue to the other conditionals like `if "tokenCount" in result:`, which will then give an error since `result` will be unbound.

You can do something like this instead:
```
if "results" in response_body and response_body["results"]:  # Check if results exists and is non-empty
      result = response_body["results"][0]
      if "tokenCount" in result:
          span.set_attribute(GEN_AI_USAGE_OUTPUT_TOKENS, result["tokenCount"])
      if "completionReason" in result:
          ...
```

Jeel-mehta: I fixed this

yiyuan-he: Looks good. Thanks!

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -240,3 +254,168 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+                        self._extract_llama_attributes(attributes, request_body)
+                    elif "cohere.command" in model_id:
+                        self._extract_cohere_attributes(attributes, request_body)
+                    elif "ai21.jamba" in model_id:
+                        self._extract_ai21_attributes(attributes, request_body)
+                    elif "mistral" in model_id:
+                        self._extract_mistral_attributes(attributes, request_body)
+
+                except json.JSONDecodeError:
+                    print("Error: Unable to parse the body as JSON")
```
mxiamxia, line 278: Can you use `_logger` for logging and change it to `debug` level? Typically for SDK implementations, we don't want to log every msg by default. It should be in `debug` level and allow customers to opt-in to enable. cc @yiyuan-he

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_bedrock_patches.py
```diff
@@ -240,3 +254,168 @@ def extract_attributes(self, attributes: _AttributeMapT):
...
+                elif "ai21.jamba" in model_id:
+                    self._handle_ai21_jamba_response(span, response_body)
+                elif "mistral" in model_id:
+                    self._handle_mistral_mistral_response(span, response_body)
+                # Replenish stream for downstream application use
+                new_stream = io.BytesIO(body_content)
+                result["body"] = StreamingBody(new_stream, len(body_content))
+
+            except json.JSONDecodeError:
+                print("Error: Unable to parse the response body as JSON")
```
mxiamxia, line 355: ditto for all `print(` in this PR.

Jeel-mehta: I made the necessary changes and made another commit

mxiamxia (approved):

PR #324, opened 2025-02-10
SigV4 Authentication support for http/protobuf exporter

liustve:
*Issue #, if available:*
Adding SigV4 Authentication extension for Exporting traces to OTLP CloudWatch endpoint without needing to explictily install the collector.

*Description of changes:*
Added a new class that extends upstream's OTLP http span exporter. Overrides the _export method so that if the endpoint is CW, we add an extra step of injecting SigV4 authentication to the headers.


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_sigv4_exporter.py
```diff
@@ -0,0 +1,118 @@
...
+                _logger.error("Failed to get credentials to export span to OTLP CloudWatch endpoint")
+
+        return super()._export(serialized_data)
+
+    def _validate_exporter_endpoint(self, endpoint: str) -> Optional[str]:
+        if not endpoint:
+            return None
+
+        region = endpoint.split(".")[1]
+        xray_regions = self.boto_session.get_available_regions(AWS_SERVICE)
```
srprash, line 108: I believe this would entail making a remote call to XRay? Even if it doesn't we don't know if the OTLP endpoint is available in all the regions where XRay is.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_sigv4_exporter.py
```diff
@@ -0,0 +1,98 @@
...
+from grpc import Compression
+
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+
+AWS_SERVICE = "xray"
+
+_logger = logging.getLogger(__name__)
+
+
+class OTLPAwsSigV4Exporter(OTLPSpanExporter):
```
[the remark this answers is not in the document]
srprash: Also, I would suggest adding some concise comments through the code explaining how the exporter works in signing the request with sigv4.

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_otlp_sigv4_exporter.py
```diff
@@ -0,0 +1,301 @@
...
+
+        mock_session.get_available_regions.assert_called_once_with("xray")
+
+    @patch.dict("sys.modules", {"botocore": None})
+    def test_no_botocore_valid_xray_endpoint(self):
+        """Test that exporter defaults when using OTLP CW endpoint without botocore"""
+
+        exporter = OTLPAwsSigV4Exporter(endpoint=OTLP_CW_ENDPOINT)
+
+        self.assertIsNone(exporter._aws_region)
```
srprash, line 86: We should be testing that the fallback exporter created is actually the `OTLPSpanExporter`

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_otlp_sigv4_exporter.py
```diff
@@ -0,0 +1,301 @@
...
+    DEFAULT_TIMEOUT,
+    DEFAULT_TRACES_EXPORT_PATH,
+    OTLPSpanExporter,
+)
+from opentelemetry.exporter.otlp.proto.http.version import __version__
+from opentelemetry.sdk.environment_variables import OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+from opentelemetry.sdk.trace import SpanContext, _Span
+from opentelemetry.trace import SpanKind, TraceFlags
+
+OTLP_CW_ENDPOINT = "https://xray.us-east-1.amazonaws.com/v1/traces"
```
majanjua-amzn, line 23: nit: Rename this accordingly as well (and where it's used), e.g.:
```suggestion
OTLP_XRAY_ENDPOINT = "https://xray.us-east-1.amazonaws.com/v1/traces"
```

aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_otlp_sigv4_exporter.py
```diff
@@ -0,0 +1,301 @@
...
+    def setUp(self):
+        self.testing_spans = [
+            self.create_span("test_span1", SpanKind.INTERNAL),
+            self.create_span("test_span2", SpanKind.SERVER),
+            self.create_span("test_span3", SpanKind.CLIENT),
+            self.create_span("test_span4", SpanKind.PRODUCER),
+            self.create_span("test_span5", SpanKind.CONSUMER),
+        ]
+
+        self.invalid_cw_otlp_tracing_endpoints = [
```
majanjua-amzn, line 41: nit:
```suggestion
        self.invalid_otlp_tracing_endpoints = [
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
```diff
@@ -0,0 +1,94 @@
...
+                method="POST",
+                url=self._endpoint,
+                data=serialized_data,
+                headers={"Content-Type": "application/x-protobuf"},
+            )
+
+            credentials = self.boto_session.get_credentials()
+
+            if credentials is not None:
+                signer = self.boto_auth.SigV4Auth(credentials, AWS_SERVICE, self._aws_region)
```
majanjua-amzn, line 82: It seems we're initializing a signer each time we export, can we make one signer object as an attribute of the class instead?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
```diff
@@ -0,0 +1,94 @@
...
+        compression: Optional[Compression] = None,
+        rsession: Optional[requests.Session] = None,
+    ):
+
+        # Represents the region of the CloudWatch OTLP endpoint to send the traces to.
+        # If the endpoint has been verified to be valid, this should not be None
+
+        self._aws_region = None
+
+        if endpoint and is_xray_otlp_endpoint(endpoint):
```
majanjua-amzn, line 42: Why do we need to check `is_xray_otlp_endpoint` again? Don't we check that before using this version of the exporter in the first place?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
majanjua-amzn, line 1: Why were the changes in this file necessary? Seems it's functionally the same

liustve: Moving it to the more general _utils python file since I also need it in the exporter.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
liustve: We do but I just think it's good to verify the structure of the endpoint before I parse it to get the region in case someone might need to manually create an instance of it.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
srprash: Curious about what happens when credentials expire? How will the credentials refresh if we only fetch them and initialize the signer only once?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
```diff
@@ -0,0 +1,94 @@
...
+            credentials = self.boto_session.get_credentials()
+
+            if credentials is not None:
+                signer = self.boto_auth.SigV4Auth(credentials, AWS_SERVICE, self._aws_region)
+
+                try:
+                    signer.add_auth(request)
+                    self._session.headers.update(dict(request.headers))
+
+                except self.boto_auth.NoCredentialsError as signing_error:
```
srprash, line 88: Can we catch the broader exception as well? I'm afraid if we are unable to handle some unexpected error and it breaks the application or the instrumentation.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
```diff
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0
 # Modifications Copyright The OpenTelemetry Authors. Licensed under the Apache License 2.0 License.
 import os
-import sys
 from logging import Logger, getLogger
 
+# pylint: disable=unused-import
+# flake8: noqa: F401
```
srprash, line 8: Why do we need these? Can we remove the `import pkg_resources`?

liustve: Yep apparently this is needed in instrumentation patch:

https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_instrumentation_patch.py#L42

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
liustve: Not entirely sure if they all share a broader exception, just try catching a general Exception

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_instrumentation_patch.py
srprash: Can we not update the import in the test to address this?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -437,6 +449,15 @@ def _is_lambda_environment():
     return AWS_LAMBDA_FUNCTION_NAME_CONFIG in os.environ
 
 
+def is_xray_otlp_endpoint(otlp_endpoint: str = None) -> bool:
+    """Is the given endpoint the XRay OTLP endpoint?"""
+
+    if not otlp_endpoint:
+        return False
+
+    return bool(re.match(XRAY_OTLP_ENDPOINT_PATTERN, otlp_endpoint.lower()))
```
majanjua-amzn, line 458: thought: We validate it's lowercase, but if it isn't actually lowercase would that have any impact later on since we're not keeping the lowercase version? I assume it's fine, just something to think about

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
majanjua-amzn: As discussed over slack, the current implementation is fine due to the credentials concern as noted by Prashant. Resolving the comment

majanjua-amzn (approved):
LGTM aside from open comments.

srprash (approved):
LGTM. TY!

PR #344, opened 2025-03-28
[SigV4] Better Error Handling and More Log Messages

liustve:
*Description of changes:*
Added better logging messages for using SigV4 and a more graceful way to handle the case where botocore is not installed to avoid this error message:
```AttributeError: 'OtlpAwsSpanExporter' object has no attribute 'boto_aws_request'```

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
@@ -328,7 +328,8 @@ def _customize_exporter(span_exporter: SpanExporter, resource: Resource) -> Span
 
         else:
             _logger.warning(
-                "Improper configuration see: please export/set OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf and OTEL_TRACES_EXPORTER=otlp"
+                "Improper configuration see: please export/set "
+                "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf and OTEL_TRACES_EXPORTER=otlp"
             )
 
     if not _is_application_signals_enabled():
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
@@ -321,13 +321,11 @@ def _customize_exporter(span_exporter: SpanExporter, resource: Resource) -> Span
 
     if is_xray_otlp_endpoint(os.environ.get(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)):
         # TODO: Change this url once doc writer has added a section for using SigV4 without collector
-        _logger.info(
-            "Detected using AWS OTLP XRay Endpoint."
-        )
+        _logger.info("Detected using AWS OTLP XRay Endpoint.")
 
         if isinstance(span_exporter, OTLPSpanExporter):
             span_exporter = OTLPAwsSpanExporter(endpoint=os.getenv(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT))
-        
+
         else:
             _logger.warning(
                 "Improper configuration see: please export/set OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf and OTEL_TRACES_EXPORTER=otlp"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py [hunk 1]
```diff
@@ -319,10 +319,19 @@ def _customize_exporter(span_exporter: SpanExporter, resource: Resource) -> Span
...
 
-    if isinstance(span_exporter, OTLPSpanExporter) and is_xray_otlp_endpoint(
-        os.environ.get(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
-    ):
-        span_exporter = OTLPAwsSpanExporter(endpoint=os.getenv(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT))
+    if is_xray_otlp_endpoint(os.environ.get(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)):
+        # TODO: Change this url once doc writer has added a section for using SigV4 without collector
+        _logger.info(
+            "Detected using AWS OTLP XRay Endpoint. Please ensure you've followed the documentation to properly setup your environment to export traces to XRay. See: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OpenTelemetry-Sections.html"
+        )
```
srprash, line 326: I don't think we need this log statement or can simply be "Detected using AWS OTLP XRay Endpoint". Users who want to send directly to the OTLP endpoint should discover this feature from the docs and should set up accordingly, then come to code. IMO if logs can't give specific instructions to solve a problem on the spot, they shouldn't point to a big doc for pre-configurations.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -319,10 +319,19 @@ def _customize_exporter(span_exporter: SpanExporter, resource: Resource) -> Span
...
+
+        if isinstance(span_exporter, OTLPSpanExporter):
+            span_exporter = OTLPAwsSpanExporter(endpoint=os.getenv(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT))
+        
+        else:
+            _logger.info(
```
srprash, line 332: maybe log at warning level?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -319,10 +319,19 @@ def _customize_exporter(span_exporter: SpanExporter, resource: Resource) -> Span
...
+                "Improper configuration see: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OpenTelemetry-Sections.html"
```
srprash, line 333: This is not the correct link right?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
```diff
@@ -74,23 +75,26 @@ def __init__(
     # the same except if the endpoint is an XRay OTLP endpoint, we will sign the request
     # with SigV4 in headers before sending it to the endpoint. Otherwise, we will skip signing.
     def _export(self, serialized_data: bytes):
-        request = self.boto_aws_request.AWSRequest(
-            method="POST",
-            url=self._endpoint,
-            data=serialized_data,
-            headers={"Content-Type": "application/x-protobuf"},
-        )
+        if self._has_required_dependencies:
```
srprash, line 78: I think we should fail much earlier if the dependency requirement is not met. The exporter shouldn't even be initialized in such a case and the SDK should fallback to using the default OTLP exporter with a warning message. 
Knowing that the exporter will not work as intended and still trying to export seems inefficient.

liustve: We don't really need to avoid instantiating this exporter if botocore isn't installed. 

1. This exporter is only instantiated if it passes the regex check to see if the endpoint matches the xray otlp endpoint so we already know that the intention of the user is to send traces to XRay.

2. In the case that botocore isn't installed, this exporter fallbacks to the behavior of the default exporter anyways. Both will be inefficient as they will attempt to export the spans anyways.

srprash (dismissed):

PR #356, opened 2025-04-25
feat: support custom log group and log stream destinations for otlp x…

yiyuan-he:
## What does this pull request do?
Adds support to customize log group and log stream destinations on OTLP X-Ray endpoint via headers.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py
@@ -1,6 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+import os
 from typing import Dict, Optional
 
 import requests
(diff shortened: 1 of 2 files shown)


yiyuan-he:
These changes may not be necessary if we don't want to support new custom env vars. Since we [inherit](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py#L16) from upstream's `OTLPSpanExporter` the following should work out of the box:

```
export OTEL_EXPORTER_OTLP_TRACES_HEADERS="x-aws-log-group=your-log-group-name,x-aws-log-stream=your-log-stream-name"
```

yiyuan-he (commented):
These changes may not be necessary if we don't want to support new custom env vars. Since we [inherit](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/otlp_aws_span_exporter.py#L16) from upstream's `OTLPSpanExporter` the following should work out of the box:

```
export OTEL_EXPORTER_OTLP_TRACES_HEADERS="x-aws-log-group=your-log-group-name,x-aws-log-stream=your-log-stream-name"
```

PR #361, opened 2025-05-12
Genesis LLO Support in ADOT SDK

yiyuan-he:
## What does this pull request do?
Adds support to handle LLO from third-party instrumentation SDKs in ADOT SDK.

The following SDKs are supported:
- Traceloop/Openllmetry
- OpenInference
- OpenLit

Note: OTel dependencies in ADOT SDK have been loosened as a short-term workaround to support the various conflicting dependency requirements of third-party instrumentation SDKs.

## Test plan
Built this custom ADOT SDK into various sample apps and exported the span and logs data to the OTLP X-Ray and Logs endpoint, respectively, to validate the LLO extraction and transformation to Gen AI Events.

Configurations tested:
- LangChain + Traceloop/Openllmetry
- LangChan + OpenInference
- LangChain + OpenLit
- CrewAI + Traceloop/Openllmetry
- CrewAI + OpenInference
- CrewAI + OpenLit

Environment variable configuration:
```
λ  env OTEL_METRICS_EXPORTER=none \
       OTEL_TRACES_EXPORTER=otlp \
       OTEL_LOGS_EXPORTER=otlp \
       OTEL_PYTHON_DISTRO=aws_distro \
       OTEL_PYTHON_CONFIGURATOR=aws_configurator \
       OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf \
       OTEL_EXPORTER_OTLP_LOGS_HEADERS="x-aws-log-group=test,x-aws-log-stream=default" \
       OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://xray.us-east-1.amazonaws.com/v1/traces \
       OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=https://logs.us-east-1.amazonaws.com/v1/logs \
       OTEL_RESOURCE_ATTRIBUTES="service.name=langchain-app" \
       OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED="true" \
       OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT="true" \
       OTEL_PYTHON_DISABLED_INSTRUMENTATIONS="http,sqlalchemy,psycopg2,pymysql,sqlite3,aiopg,asyncpg,mysql_connector,botocore,boto3,urllib3,requests,starlette" \
       AGENT_OBSERVABILITY_ENABLED="true" \
       python app.py
```

[...]

PR #364, opened 2025-05-18
[WIP] Support for handling LLO in OTLP AWS Logs Exporter

liustve:
Supporting ADOT auto instrumentation to handle LLO log events.

*Description of changes:*

1. Adds `AwsBatchLogRecordProcessor` a backwards compatible custom logs BatchProcessor which has the following invariants: 
    - The unserialized, uncompressed data size of exported batches will ALWAYS be <= 1 MB except for the case below:  
    - If the data size of an `exported batch is ever > 1 MB` then the batch size is always length 1
2.  `OTLPAwsLogExporter`: Adds a new behavior for Retry delay based on server-side response of Retry-After header. Injects the LLO header flag if the size of the exported data > 1 MB.
3. Customize the auto instrumentation to use the new `AwsBatchLogRecordProcessor`

*Testing:*

TODO:
1.  Add unit tests to validate behavior of `AwsBatchLogRecordProcessor` and  `OTLPAwsLogExporter`
2. E2E testing to validate any performance hits and compatibility with: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/361


yiyuan-he:
Thanks! Looks good overall.

Couple of small things:
- Could you rebase again - seems like there are some extra changes not related to the processor.
- Print statement cleanup.
- We can just add a TODO for the retry logic and raise another PR.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/otlp/aws/common/constants.py
```diff
@@ -0,0 +1,4 @@
+BASE_LOG_BUFFER_BYTE_SIZE = 2000
```
yiyuan-he, line 1: Where does this number come from?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/otlp/aws/logs/aws_batch_log_record_processor.py
```diff
@@ -0,0 +1,130 @@
...
+        if isinstance(val, int) or isinstance(val, float):
+            return len(str(val))
+
+        if isinstance(val, Sequence):
+            for content in val:
+                size += AwsBatchLogRecordProcessor._get_size_of_any_value(content)
+
+        if isinstance(val, Mapping):
+            for _, content in val.items():
+                size += AwsBatchLogRecordProcessor._get_size_of_any_value(content)
```
yiyuan-he, line 128: Do we need these? The Gen AI Events we are emitting from the Span pipeline only contain primitive data types.

This recursive approach traverses the entire object graph depth-first, which could become expensive for:
  - Deeply nested JSON structures
  - Large arrays of complex objects
  - Circular references (would cause infinite recursion if present)
  - Dictionary keys aren't counted in the size calculation (potential underestimation)

liustve: This processor will have to account for more than just gen ai events. All logs emitted by the application will be processed through this so we have to account for more than primitive types.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/otlp/aws/common/constants.py
liustve: I ran an e2e test with the Gen AI sample app and subtracted the size of the body content from the size of the entire log itself. I then multiplied by 3 and added a bit of extra buffer to account for larger logs.

PR #381, opened 2025-06-04
Drop Python 3.8 Support in ADOT SDK

yiyuan-he:
## What does this pull request do?
Removes Python 3.8 support in our ADOT SDK since it has reached 6 months past EoL, which is the duration that upstream OTel commits to supporting.

Found all references to Python 3.8 via [GitHub repo search](https://github.com/search?q=repo%3Aaws-observability%2Faws-otel-python-instrumentation%203.8&type=code).


contract-tests/tests/pyproject.toml
```diff
@@ -32,4 +32,4 @@ packages = ["test/amazon"]
 log_cli = true
 log_cli_level = "INFO"
 log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
-log_cli_date_format = "%Y-%m-%d %H:%M:%S"
\ No newline at end of file
+log_cli_date_format = "%Y-%m-%d %H:%M:%S"
```
yiyuan-he, line 35: I'm not sure why this happens - I only changed line 10 above.

PR #414, opened 2025-06-30
Improve resource attributes handling

mxiamxia:
*Description of changes:*
  - Implement automatic injection of aws.ai.agent.type resource attribute when agent observability is enabled
  - Set default value to "default" if not already present in resource
  - Add comprehensive unit tests for _customize_resource function
  - Diable default resource detectors for GenAI cases


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
```diff
@@ -117,5 +115,7 @@ def _configure(self, **kwargs):
             # Disable AWS Application Signals by default
             os.environ.setdefault(APPLICATION_SIGNALS_ENABLED_CONFIG, "false")
 
+        super(AwsOpenTelemetryDistro, self)._configure()
```
jj22ee, line 118: Just wonder if there was a reason that the super() call was made before we set most (but not all) of the environment variables.

srprash, line 118: Curious on why did we have to move this down?

srprash (approved):

jj22ee (approved):

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_distro.py
mxiamxia: >Just wonder if there was a reason that the super() call was made before we set most (but not all) of the environment variables.

It was a code merge issue. In beta version, we have super._configure() occurred after Agentic default setting.

mxiamxia: > Curious on why did we have to move this down?

`super(AwsOpenTelemetryDistro, self)._configure()` will set all OTel Exporters to `otlp` by default only if they're not set. So we want to set it first (esp for OTel Metrics Exp) and then apply this defaulting logic

PR #415, opened 2025-07-01
Rename aws.ai.agent.type to aws.service.type for agent observability

mxiamxia:
*Description of changes:*
 - Rename `aws.ai.agent.type` to/and set ` aws.service.type` with default value `gen_ai_agent` if no custom set
 - rename 


srprash (approved):
LGTM. However, I have some concerns with introducing `aws.service.type` without having thought of cases other than `gen_ai_agent`. Since it is such a broad attribute, we probably need to have an idea on possible values for it and some spec around those values.
Anyway, I know we want to get this in quickly and I don't have strong suggestions for this right now. Hopefully, this will work out well in the future.

PR #431, opened 2025-07-13
Fix Starlette auto-instrumentation and botocore propagator configuration

mxiamxia:
*Description of changes:*
   - Enable Starlette auto-instrumentation by removing it from disabled list
   - Add patch to relax Starlette version constraint (>=0.13, <0.15 to >=0.13)
   - Fix botocore instrumentation to use global propagator instead of hardcoded XRay
image


wangzlei (approved):
LGTM! Thanks!

PR #437, opened 2025-07-24
Support export OTel metrics to be EMF in starndard output

wangzlei:
*Issue #, if available:*
Currently, OpenTelemetry (OTel) metrics users in AWS Lambda must export metrics via synchronous calls. This often results in increased function duration or lost metric data, due to how Lambda handles execution lifecycle — particularly during the [freeze phase](https://serverlessfolks.com/lambda-code-execution-freezethaw#heading-freeze).

*Description of changes:*
This PR introduces a new ConsoleEmfExporter, which exports OTel metrics to standard output using the CloudWatch Embedded Metric Format (EMF).
In the Lambda environment, logs written to standard output are automatically forwarded to CloudWatch by Lambda's built-in logging agent. This makes ConsoleEmfExporter a simple and efficient way to export OTel metrics in Lambda, avoiding the overhead and reliability issues associated with synchronous metric export.


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -634,10 +634,11 @@ def _validate_and_fetch_logs_header() -> OtlpLogHeaderSetting:
...
-        _logger.warning(
-            "Improper configuration: Please configure the environment variable OTEL_EXPORTER_OTLP_LOGS_HEADERS "
-            "to include x-aws-log-group and x-aws-log-stream"
-        )
+        if not _is_lambda_environment():
+            _logger.warning(
+                "Improper configuration: Please configure the environment variable OTEL_EXPORTER_OTLP_LOGS_HEADERS "
+                "to include x-aws-log-group and x-aws-log-stream"
+            )
         return OtlpLogHeaderSetting(None, None, None, False)
```
jj22ee, line 642: Here we are saying using this header as a valid header for Lambda, but still return `is_valid=false`. I think we should get rid of `is_valid` in `OtlpLogHeaderSetting` because it is redundant since we also return values for logGroup/logStream. Wherever we use the `is_valid` value, just replace it to check that logGroup and logStream are defined.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py
```diff
@@ -771,8 +772,24 @@ def _check_emf_exporter_enabled() -> bool:
 
     return True
 
-
 def _create_emf_exporter():
+    """
+    Create the appropriate EMF exporter based on the environment and configuration.
+
+    """
+    log_header_setting = _validate_and_fetch_logs_header()
```
jj22ee, line 780: `_validate_and_fetch_logs_header()` can be ran again when `_create_cloudwatch_emf_exporter()` executed. Maybe we can update `_create_cloudwatch_emf_exporter` to take in an optional value for `log_header_setting`? e.g. if we provide `log_header_setting` to `_create_cloudwatch_emf_exporter`, then it won't re-fetch the log headers. If not provided, use existing logic.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/aws/metrics/console_emf_exporter.py
```diff
@@ -0,0 +1,107 @@
...
+            # Write the EMF log message to stdout for easy redirection/capture
+            message = log_event.get("message", "")
+            if message:
+                print(message, flush=True)
+            else:
+                logger.warning("Empty message in log event: %s", log_event)
+        except (KeyError, TypeError) as e:
+            logger.error("Invalid log event format: %s. Error: %s", log_event, e)
+        except Exception as e:
+            logger.error("Failed to write EMF log to console: %s", e)
```
jj22ee, line 74: Are these exceptions possible for what is done in the `try`?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/aws/metrics/console_emf_exporter.py
```diff
@@ -0,0 +1,107 @@
...
+            **kwargs: Additional arguments (unused, kept for compatibility)
+        """
+        # No need to check for None since namespace has a default value
+        if namespace is None:
+            namespace = "default"
+        super().__init__(namespace, preferred_temporality, preferred_aggregation)
+
+    def _send_log_event(self, log_event: Dict[str, Any]) -> None:
+        """
+        Send a log event to stdout for console output.
```
jj22ee, line 52:
```suggestion
        Send a log event message to stdout for console output.
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/aws/metrics/console_emf_exporter.py
```diff
@@ -0,0 +1,107 @@
...
+            preferred_temporality: Optional dictionary mapping instrument types to aggregation temporality
+            preferred_aggregation: Optional dictionary mapping instrument types to preferred aggregation
+            **kwargs: Additional arguments (unused, kept for compatibility)
+        """
+        # No need to check for None since namespace has a default value
+        if namespace is None:
+            namespace = "default"
+        super().__init__(namespace, preferred_temporality, preferred_aggregation)
+
+    def _send_log_event(self, log_event: Dict[str, Any]) -> None:
```
jj22ee, line 50: Since `_send_log_event` is more `aws_cloudwatch_emf_exporter` specific as we don't send LogEvent in console exporter, could this method name be made more generic? Maybe `_export`?

mxiamxia (approved):

PR #689, opened 2026-03-26
Fix: _send_log_batch fails permanently when log group is managed by IaC

thenom:
Fixes #688

# Problem

When PutLogEvents fails with ResourceNotFoundException, the recovery path in _send_log_batch unconditionally calls _create_log_group_if_needed() before _create_log_stream_if_needed(). If the IAM role lacks logs:CreateLogGroup (because the log group is pre-created by Terraform/IaC), the AccessDeniedException propagates and CreateLogStream is never attempted. The EMF metrics exporter then fails permanently, logging errors every ~60s.

# Fix

Reorder the recovery path to try creating the log stream first. If that succeeds (log group already exists), log group creation is skipped entirely. Only if CreateLogStream itself fails with ResourceNotFoundException (meaning the group genuinely doesn't exist) does it fall back to creating both.

# Changes

_cloudwatch_log_client.py: Reordered resource creation in _send_log_batch — stream-first, group only if needed.
test_cloudwatch_log_client.py: Updated existing tests to reflect stream-first ordering, added tests for the IaC-managed log group scenario and the both-missing scenario.


viw-test1:
Update:

1. Thanks for the report and the PR. The analysis makes sense.
2. The bug is when the log group is pre-created by IaC and the role lacks logs:CreateLogGroup, the AccessDeniedException from CreateLogGroup prevents CreateLogStream from ever being attempted.
3. I'll review the suggested PR #689.

viw-test1 (approved):
Looks good to me

srprash (approved):

PR #694, opened 2026-03-27
Add OpenTelemetry version compatibility check at distro startup

wangzlei:
## Summary
- Add early version compatibility check for `opentelemetry-api` and `opentelemetry-sdk` at distro startup
- Compares installed versions against expected versions from `aws-opentelemetry-distro` package metadata
- Logs a warning when mismatch is detected, before import errors can occur
- Helps users diagnose version conflicts in Lambda layer deployments where user code bundles a different OTel version

## Design decisions
- Runs before any `opentelemetry` imports to catch mismatches early
- Uses `importlib.metadata` (stdlib) with simple string parsing instead of `packaging.requirements` to minimize Lambda cold start overhead (~15ms vs ~50ms)
- Best-effort check wrapped in try/except — never blocks startup

## Test result

```
(.venv) [/Volumes/workplace/release/python-compatible]$ opentelemetry-instrument python sample.py
OpenTelemetry package version mismatch: opentelemetry-api==1.33.1, opentelemetry-sdk==1.39.1. AWS OpenTelemetry Distro expects opentelemetry-api==1.40.0, opentelemetry-sdk==1.40.0, which may cause unexpected errors.
```


vastin (dismissed):
LGTM

vastin (approved):

PR #715, opened 2026-04-09
Standardize Python Log Exporter JSON output to match canonical schema

Miqueasher:
The ADOT Lambda layers support `OTEL_LOGS_EXPORTER=console`, which writes
log records as compact JSON to stdout. Lambda's FluxPump reads stdout and
forwards to PLE. Previously, each language (Java, Python, JS, .NET) produced
different JSON output because Python and JS delegated to their upstream OTel
SDK's serialization methods, which each made independent formatting choices
(snake_case vs camelCase field names, 0x-prefixed trace IDs, different
timestamp formats, missing fields).

This PR aligns the Python exporter output with the canonical schema defined
by the Java implementation, ensuring consistent JSON structure across all
ADOT language SDKs

- Rewrites the `CompactConsoleLogRecordExporter` to build JSON directly from
  `LogRecord` fields instead of delegating to the upstream SDK's `to_json()`
  serialization, which produced a different schema than the Java reference

- Adds try-catch fallback to the original upstream SDK format if the new
  serialization fails, to avoid breaking existing customer infrastructure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- -fix(lambda-layer): Standardize CompactConsoleLogRecordExporter output with CloudWatch OTLP backend schema.
+  ([#715](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/715))
 - fix(lambda-layer): Disable all agentic instrumentation in Lambda by default
   ([#710](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/710))
 - fix(genai-instrumentors): cleanup code, align with OTel GenAI semconv, add missing attributes and fix deprecated usage
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
-- -fix(lambda-layer): Standardize CompactConsoleLogRecordExporter output with CloudWatch OTLP backend schema.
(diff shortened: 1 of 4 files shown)


wangzlei (commented):
~~Superseded by updated review below.~~

wangzlei (commented):
## Review Comments 

Thanks for the work on standardizing the console exporter output! Two items need to be fixed:

### 1. `LogExportResult` import missing fallback 

The base class import has a try/except for `LogRecordExporter` vs `LogExporter`, but the return type import doesn't:

```python
from opentelemetry.sdk._logs.export import LogExportResult  # no fallback
```

Older SDK versions export `LogRecordExportResult`, not `LogExportResult`. This will crash at import time on those versions. Should add:

```python
try:
    from opentelemetry.sdk._logs.export import LogExportResult
except ImportError:
    from opentelemetry.sdk._logs.export import LogRecordExportResult as LogExportResult
```

Same issue in the test file.

### 2. Boolean attribute values: `"True"` vs `"true"` 

```python
attrs[k] = str(v)  # Python str(True) → "True"
```

Java's `String.valueOf(true)` produces `"true"`. Since the goal is matching the canonical schema across all languages, this is a mismatch. The test asserts `"True"` which confirms the divergence.

Suggest: `str(v).lower() if isinstance(v, bool) else str(v)`, or a more explicit type coercion.

liustve:
consider using the JSON formatter in here: https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/_logs/_internal/__init__.py#L214

wangzlei (dismissed):

wangzlei (approved):

PR #727, opened 2026-04-21
fix(lambda): align Python context propagation with JS implementation

wangzlei:
Closes https://github.com/aws-observability/aws-otel-python-instrumentation/issues/663

## Summary
- **Problem:** `custom_event_context_extractor` in `otel_wrapper.py` always used the X-Ray env var (`_X_AMZN_TRACE_ID`) as parent context because it is always set and valid when active tracing is enabled, causing upstream W3C trace context (`traceparent`) from event headers to be ignored and breaking trace continuity.
- **Fix:** Align with the [JS implementation](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts#L100-L123) — inject the X-Ray env var into the event headers (replacing any existing X-Ray header, case-insensitive), then delegate to the global composite propagator via `get_global_textmap().extract()`. This respects the propagator priority ordering configured in `otel-instrument` (`baggage,xray,tracecontext`), where `tracecontext` (W3C) takes precedence over `xray` when both are present.
- **Tests updated:** Existing tests adapted to explicitly set the global textmap propagator. Two new tests added:
  - `test_xray_ignored_when_propagator_does_not_include_xray` — X-Ray active tracing on, but only `tracecontext` propagator configured; W3C headers should be used.
  - `test_w3c_takes_precedence_over_xray_when_both_present` — both X-Ray and W3C headers present with composite propagator; W3C should win.

## Test plan
- [x] All 4 unit tests pass locally (`pytest -v`)
- [x] Coverage: `otel_wrapper.py` at 90% (uncovered lines are error-handling edge cases)
- [x] Lint passes: `black`, `isort`, `flake8` all clean

### DIFF
--- a/lambda-layer/src/otel_wrapper.py
+++ b/lambda-layer/src/otel_wrapper.py
@@ -41,7 +41,6 @@ from typing import Any
 from opentelemetry.context import Context
 from opentelemetry.instrumentation.aws_lambda import _X_AMZN_TRACE_ID, AwsLambdaInstrumentor
 from opentelemetry.propagate import get_global_textmap
-from opentelemetry.propagators.aws import AwsXRayPropagator
 from opentelemetry.propagators.aws.aws_xray_propagator import TRACE_HEADER_KEY
 from opentelemetry.trace import get_current_span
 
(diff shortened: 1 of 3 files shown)


lambda-layer/src/otel_wrapper.py
```diff
@@ -57,21 +56,24 @@ class HandlerError(Exception):
...
-        try:
-            headers = lambda_event["headers"]
-        except (TypeError, KeyError):
-            pass
-        if not isinstance(headers, dict):
-            headers = {}
+    if not isinstance(headers, dict):
+        headers = {}
+    else:
+        headers = {k: v for k, v in headers.items()}
```
jj22ee: nit - If this is so we create a shallow copy(?), `headers = headers.copy()` may be more pythonic.

jj22ee (dismissed):

vastin (dismissed):

vastin (approved):

PR #744, opened 2026-05-12
fix AttributePropagatingSpanProcessor with NonRecordingSpan as parent

### LINKED ISSUE #743: AttributePropagatingSpanProcessor doesn't accept non-Readable spans
Hi,

I'm encountering an error when creating an Otel span with ADOT

https://github.com/aws-observability/aws-otel-python-instrumentation/blob/28fc34eb3f0a1e11d6328cdee55b7eaba501d2af/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/attribute_propagating_span_processor.py#L80-L81

When the `parent_span` isn't a `ReadableSpan` (e.g. a `NonRecordingSpan`), `_is_server_kind` would throw an attribute exception because the other types of span may not have the `kind` attribute.

My use case: 

I'm using `NonRecordingSpan` to create a parent span placeholder, which will be filled later

```
        context = SpanContext(
            trace_id=self._id_generator.generate_trace_id(),
            span_id=operation_id_to_span_id(parent_id),
            is_remote=False,
        )
        placeholder_span = NonRecordingSpan(context)

        parent_context = trace.set_span_in_context(
                placeholder_span, self._extracted_context
            )

        span = self._tracer.start_span(
            name=name,
            attributes=attributes,
            start_time=_to_otel_timestamp(start_time),
            context=parent_context,
        )
```

My workaround is to add the expected attributes to the `NonRecordingSpan` I created above:

```
        placeholder_span.kind = SpanKind.INTERNAL
        placeholder_span.attributes = {}
```

zhongkechen:
*Issue #, if available:*

Fixes: #743 

*Description of changes:*

https://github.com/aws-observability/aws-otel-python-instrumentation/blob/28fc34eb3f0a1e11d6328cdee55b7eaba501d2af/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/attribute_propagating_span_processor.py#L80-L81

When the `parent_span` isn't a `ReadableSpan` (e.g. a `NonRecordingSpan`), `_is_server_kind` would throw an attribute exception because the other types of span may not have the `kind` attribute.


liustve:
I partially am okay with these changes, though I don't agree with this span processor working with NonRecording Spans since there's no SpanKind. 
https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L523

liustve (commented):
instead of `isinstance(parent_span, ReadableSpan)`, could we use `parent_span.is_recording()` instead?

The OTel spec [defines](https://opentelemetry.io/docs/specs/otel/trace/api/#wrapping-a-spancontext-in-a-span) that `NonRecordingSpan.is_recording()` must return `False`, so it's a cleaner check that doesn't require importing `ReadableSpan` and aligns directly with the spec's contract:

```suggestion
        elif parent_span and parent_span.is_recording() and _is_server_kind(parent_span):
            propagation_data = self._propagation_data_extractor(parent_span)
        elif parent_span and parent_span.is_recording():
            propagation_data = parent_span.attributes.get(self._propagation_data_key)
```

liustve:
@zhongkechen Please add unit tests for this change as well and a `CHANGELOG.md` entry

zhongkechen:
> instead of `isinstance(parent_span, ReadableSpan)`, could we use `parent_span.is_recording()` instead?
> 
> The OTel spec [defines](https://opentelemetry.io/docs/specs/otel/trace/api/#wrapping-a-spancontext-in-a-span) that `NonRecordingSpan.is_recording()` must return `False`, so it's a cleaner check that doesn't require importing `ReadableSpan` and aligns directly with the spec's contract:
> 
> ```
>         elif parent_span and parent_span.is_recording() and _is_server_kind(parent_span):
>             propagation_data = self._propagation_data_extractor(parent_span)
>         elif parent_span and parent_span.is_recording():
>             propagation_data = parent_span.attributes.get(self._propagation_data_key)
> ```

I'm using `isinstance(parent_span, ReadableSpan)` because a few lines above my change is using that:
 
https://github.com/zhongkechen/aws-otel-python-instrumentation/blob/0ac556982b8a06e456018e3dbd2948898dfe96e7/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/attribute_propagating_span_processor.py#L50

```
    def on_start(self, span: Span, parent_context: Optional[Context] = None) -> None:
        parent_span: ReadableSpan = get_current_span(parent_context)
        if isinstance(parent_span, ReadableSpan):
```

I thought I should keep the check logic consistent at least in the same method

liustve:
> > instead of `isinstance(parent_span, ReadableSpan)`, could we use `parent_span.is_recording()` instead?
> > The OTel spec [defines](https://opentelemetry.io/docs/specs/otel/trace/api/#wrapping-a-spancontext-in-a-span) that `NonRecordingSpan.is_recording()` must return `False`, so it's a cleaner check that doesn't require importing `ReadableSpan` and aligns directly with the spec's contract:
> > ```
> >         elif parent_span and parent_span.is_recording() and _is_server_kind(parent_span):
> >             propagation_data = self._propagation_data_extractor(parent_span)
> >         elif parent_span and parent_span.is_recording():
> >             propagation_data = parent_span.attributes.get(self._propagation_data_key)
> > ```
> 
> I'm using `isinstance(parent_span, ReadableSpan)` because a few lines above my change is using that:
> 
> https://github.com/zhongkechen/aws-otel-python-instrumentation/blob/0ac556982b8a06e456018e3dbd2948898dfe96e7/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/attribute_propagating_span_processor.py#L50
> 
> ```
>     def on_start(self, span: Span, parent_context: Optional[Context] = None) -> None:
>         parent_span: ReadableSpan = get_current_span(parent_context)
>         if isinstance(parent_span, ReadableSpan):
> ```
> 
> I thought I should keep the check logic consistent at least in the same method


I see though I'd still suggest using what is OTel spec aligned instead of using `isinstance` imo.

Aneurysm9:
I don't think `parent_span.is_recording()` actually conveys the necessary information here.  That's an [API method](https://opentelemetry.io/docs/specs/otel/trace/api/#isrecording) intended for allowing instrumentation to skip potentially expensive attribute calculation/creation operations if the result won't be recorded.  An API span may be recording and still write-only.  Also, `OnStart()` should only ever be called on recording spans and I'm not aware of any situation in which a recording span would have a non-recording parent span.  Maybe if the parent span was ended before the child was started, but even then it may be possible to get a readable span from the parent despite it no longer being in a recording state.

Since this is an [SDK plugin](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor) it should be fine for it to introspect API spans to get at the underlying SDK spans, which is where `ReadableSpan` would fit in.  The thing I'm not sure of with the Python implementation is whether a `ReadableSpan` is always read-only, or if a read/write span also satisfies `isinstance(ps, ReadableSpan)`.

liustve:
> I don't think `parent_span.is_recording()` actually conveys the necessary information here. That's an [API method](https://opentelemetry.io/docs/specs/otel/trace/api/#isrecording) intended for allowing instrumentation to skip potentially expensive attribute calculation/creation operations if the result won't be recorded. An API span may be recording and still write-only. Also, `OnStart()` should only ever be called on recording spans and I'm not aware of any situation in which a recording span would have a non-recording parent span. Maybe if the parent span was ended before the child was started, but even then it may be possible to get a readable span from the parent despite it no longer being in a recording state.
> 
> Since this is an [SDK plugin](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor) it should be fine for it to introspect API spans to get at the underlying SDK spans, which is where `ReadableSpan` would fit in. The thing I'm not sure of with the Python implementation is whether a `ReadableSpan` is always read-only, or if a read/write span also satisfies `isinstance(ps, ReadableSpan)`.

I'm okay with the changes then, @zhongkechen please add unit tests and a changelog entry and we can help merge this

zhongkechen:
added a unit test case

liustve (approved):

srprash (approved):

PR #749, opened 2026-05-21
fix(otlp-aws-exporter): avoid RecursionError with pip_system_certs on…

v-iw:
… Python 3.12

*Issue #, if available:*
*Description of changes:*

Problem:
When pip_system_certs is installed alongside ADOT Python auto-instrumentation, the OTLP AWS exporter can fail with a RecursionError while setting up the TLS connection for an export, and the same failure can surface in application code that uses boto3/botocore. The effect is that exports fail repeatedly and no telemetry reaches CloudWatch.

Root cause:
pip_system_certs enables OS/corporate trust store support by calling truststore.inject_into_ssl(), which replaces ssl.SSLContext process-wide with the truststore-backed class. It does this from a .pth file at interpreter startup, in the finally block of a site.execsitecustomize wrapper — i.e. after sitecustomize.py returns.

ADOT's auto-instrumentation runs from sitecustomize.py (via opentelemetry-instrument). During that phase it transitively imports requests (through the upstream OTLP HTTP exporters) and botocore. Both capture a reference to ssl.SSLContext at import time — and because that happens before truststore's injection, they capture the original ssl.SSLContext rather than the truststore-backed class.

On affected CPython versions, the ssl.SSLContext.options setter resolves SSLContext from the ssl module's globals at call time. Once truststore has been injected, that name resolves to the truststore-backed class, and the super() chain in the setter recurses between the two classes until the recursion limit is exceeded.

The OTLP AWS exporter is affected because it signs each export with SigV4, and the credential resolver constructs an SSL context as part of that path — so the failure recurs on every export.

Fix:
1. Add _pip_system_certs_patches.py
 with a one-shot, lock-guarded helper that rebinds botocore.httpsession.SSLContext and urllib3.util.ssl_.SSLContext to the current ssl.SSLContext (the truststore-backed class). That class's options setter does not use the recursive super() pattern, so SSL context construction succeeds. The helper is a no-op when pip_system_certs is not installed.

2. Apply the patch on the first AwsAuthSession.request() call. This is the earliest point at which the deferred trust store injection is guaranteed to have completed, so the rebind targets the correct class.

3. Cache the resolved AWS credentials on first use in AwsAuthSession. The credential resolver also constructs an SSL context, so resolving once per exporter (rather than once per export) limits SSL context construction. RefreshableCredentials refreshes internally on attribute access, so caching the reference does not cache stale values. On a transient resolution failure the cache is not latched, so the next request retries.

Testing:
1. New test_pip_system_certs_patches.py covers the no-op (not installed), rebind, already-matching, import-missing, and one-shot guard paths.

2. test_aws_auth_session.py adds assertions that credentials are resolved once across multiple requests, that resolution retries after a transient failure, that the compatibility patch is invoked once, and that credential failures are logged exactly once.

3. Manually validated by reproducing the failure against a published release and confirming it no longer occurs with this change, under matching conditions (pip_system_certs installed, trust store injected after botocore/urllib3 captured the original ssl.SSLContext).


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/exporter/otlp/aws/common/aws_auth_session.py
```diff
@@ -39,13 +44,57 @@ def __init__(self, aws_region: str, service: str, session: Session):
...
+                )
+
+            try:
+                self._credentials = self._session.get_credentials()
+            except Exception as cred_error:  # pylint: disable=broad-except
+                _logger.error("Failed to load AWS Credentials: %s", cred_error)
+                self._credentials = None
+
+            self._credentials_resolved = True
+
```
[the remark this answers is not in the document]
ezhang6811: @viw-test1 can you address this?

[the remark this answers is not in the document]
v-iw: This is an old comment and it has been resolved already.

srprash:
Waiting until the first export and then applying the rebind for the stale `ssl.SSLContext` seems alright, but it still leaves the gap open for other paths in the application where a boto3 client may be created with the stale class before the first export could fire. I suggest we look into plugging that gap too.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- fix(otlp-aws-exporter): avoid `RecursionError` when `pip_system_certs` replaces `ssl.SSLContext` on Python 3.12 by rebinding stale `botocore`/`urllib3` SSL context references and caching credentials in `AwsAuthSession`
```
srprash: not only limited to py3.12. please update this appropriately.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/patches/_pip_system_certs_patches.py
```diff
@@ -0,0 +1,113 @@
...
+
+    with _patch_lock:
+        # Double-checked locking: another thread may have completed the patch
+        # while we were waiting on the lock.
+        if _patch_applied:
+            return
+
+        # Only apply the patch when pip_system_certs is installed in user application space.
+        if not _is_pip_system_certs_installed():
+            _patch_applied = True
```
srprash: Should this `_patch_applied` still be set to true even if we did not apply it?

v-iw: The name was mislead but the logic is correct. The reason we still have `_patch_applied` there is to avoid repeating the importlib.metadata.version() lookup on every export. I've renamed the flag to _patch_attempted to reflect that it records "the one-time check ran," and added a comment explaining why we latch it even when there's nothing to patch.

srprash (approved):

PR #759, opened 2026-05-29
fix: Tag platform images via ECR API to fix wrapper bug in private ECR

wangzlei:
## Summary

- Removes the two single-platform build/push steps that produced wrapper-index tags instead of real platform manifests
- Keeps the single multi-arch `docker/build-push-action` push to both private and public ECR
- Adds an ECR-API retag step that reads the multi-arch manifest index and applies `v$VERSION-amd64` / `v$VERSION-arm64` directly to the platform digests via `aws ecr put-image`
- Tags attestation manifests (architecture=unknown) with their sha hex

This is the same fix applied in:
- Java: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1377
- JS: https://github.com/aws-observability/aws-otel-js-instrumentation/pull/456

## Test plan

- [ ] Trigger a release build and verify `v$VERSION-amd64` / `v$VERSION-arm64` tags in private ECR point to platform manifests (not wrapper indexes)
- [ ] Verify `v$VERSION` multi-arch tag still works correctly in both public and private ECR
- [ ] Verify attestation manifests are tagged with their sha hex


vastin (approved):

v-iw (approved):
good

PR #763, opened 2026-06-09
feat(serviceevents): add ServiceEvents feature

vastin:
## Description of changes

This PR adds the **ServiceEvents** feature to the ADOT Python distro: a set of
higher-level telemetry signals derived from function-level and HTTP endpoint
activity, emitted via OTLP.

ServiceEvents is **opt-in and bundled-gated behind Application Signals**
(`OTEL_AWS_SERVICE_EVENTS_ENABLED` / `OTEL_AWS_APPLICATION_SIGNALS_ENABLED`); it
is disabled in AWS Lambda.

## Summary

When active, ServiceEvents emits function-call duration metrics, HTTP endpoint
summaries, per-exception error counts, deployment events, and incident snapshots.
Function-level instrumentation is scoped by an explicit allowlist, and endpoint
tracking is wired through framework middleware for Flask, FastAPI, and Django.

## What's included

- ServiceEvents package and configurator wiring to initialize it when enabled.
- Unit tests for the package.
- Contract tests with sample apps and mock-collector infrastructure.

## Testing

- Unit and contract test suites pass.
- `tox -e lint` and `tox -e spellcheck` pass.

By submitting this pull request, I confirm that my contribution is made under

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
@@ -34,6 +34,22 @@ _registry_lock = threading.Lock()
 FunctionDef = TypeVar("FunctionDef", ast.FunctionDef, ast.AsyncFunctionDef)
 
 
+def _is_docstring_expr(node: ast.stmt) -> bool:
+    """Return True if ``node`` is a string-constant expression statement (a docstring).
+
+    A docstring is only recognised as the FIRST statement of a module/function body. Handles
+    the 3.8 AST change (``ast.Str`` with ``.s`` → ``ast.Constant`` with ``.value``); ``ast.Str``
+    is deprecated and removed in newer CPython, so the version gate keeps this working across
+    interpreters. Shared by the module-docstring check (preamble insertion) and the
+    function-docstring check (body rewrite) so the two cannot drift apart.
+    """
+    if not isinstance(node, ast.Expr):
+        return False
+    if sys.version_info >= (3, 8):
+        return isinstance(node.value, ast.Constant) and isinstance(node.value.value, str)
+    return isinstance(node.value, ast.Str) and isinstance(node.value.s, str)  # type: ignore[attr-defined]
+
+
 def _file_path_to_module_path(file_path: str) -> str:
     """Convert a file path to a module-style path for function naming.
 
@@ -226,21 +242,7 @@ class ServiceEventsASTTransformer(ast.NodeTransformer):
         node: Union[ast.FunctionDef, ast.AsyncFunctionDef],
     ) -> Optional[ast.stmt]:
         """If the first expression in the function is a docstring, remove and return it."""
-        # Mirrors an ast type name (Constant/Str); kept CamelCase to match the ast API.
-        AstStrType = ast.Constant if sys.version_info >= (3, 8) else ast.Str  # pylint: disable=invalid-name
-
-        if not node.body:
-            return None
-
-        if (
-            isinstance(node.body[0], ast.Expr)
-            and isinstance(node.body[0].value, AstStrType)
-            and (
-                isinstance(node.body[0].value.value, str)  # type: ignore[attr-defined]
-                if sys.version_info >= (3, 8)
-                else isinstance(node.body[0].value.s, str)  # type: ignore[attr-defined]
-            )
-        ):
+        if node.body and _is_docstring_expr(node.body[0]):
             return node.body.pop(0)
         return None
 
@@ -407,6 +409,35 @@ class ServiceEventsSourceLoader(SourceFileLoader):
         # Fallback: return the original path unchanged
         return source_path
 
+    @staticmethod
+    def _preamble_insert_index(body: List[ast.stmt]) -> int:
+        """Index at which an injected import may go without displacing required leading statements.
+
+        Returns the position just past the module's mandatory "preamble" so the injected
+        ``PythonServiceEventsMonitor`` import cannot push aside statements Python requires
+        to come first:
+          1. A module docstring — ``__doc__`` is populated only when the FIRST statement is a
+             string-constant ``Expr``; inserting before it silently makes ``__doc__`` None
+             (breaks pydoc/help()/doctest/``__doc__``-based tooling).
+          2. ``from __future__ import ...`` — these MUST be the first statements (after any
+             docstring); inserting before one raises ``SyntaxError`` at compile(), which
+             get_code's broad-except would swallow into a silent fall back to *uninstrumented*
+             loading for the whole module.
+
+        Note on (2): in the current pipeline ServiceEventsASTTransformer.visit_ImportFrom strips
+        every ``__future__`` node into compiler flags BEFORE this runs, so on a transformed tree
+        no ``__future__`` node survives for the loop to skip. The skip is kept deliberately so this
+        helper is correct on ANY body (it is unit-tested directly on un-transformed bodies, and
+        stays safe if that strip step is ever changed) — not because the loop fires today.
+        """
+        if not body:
+            return 0
+        index = 1 if _is_docstring_expr(body[0]) else 0
+        # Skip past any leading ``from __future__`` imports (they may follow the docstring).
+        while index < len(body) and isinstance(body[index], ast.ImportFrom) and body[index].module == "__future__":
+            index += 1
+        return index
+
     def get_code(self, fullname: str) -> Optional[types.CodeType]:
         """Load and transform the module's source code."""
         source_path = self.get_filename(fullname)
@@ -440,22 +471,10 @@ class ServiceEventsSourceLoader(SourceFileLoader):
                 level=0,
             )
             ast.fix_missing_locations(import_node)
-            # Insert the import after a leading module docstring (if any). A module's
-            # __doc__ is populated only when the FIRST statement is a string-constant Expr;
-            # inserting the import at index 0 would displace the docstring and silently make
-            # __doc__ None, breaking pydoc/help()/doctest/__doc__-based tooling. Mirrors the
-            # docstring predicate in get_and_remove_docstring (incl. the 3.8 ast.Str fallback).
-            AstStrType = ast.Constant if sys.version_info >= (3, 8) else ast.Str  # pylint: disable=invalid-name
-            has_leading_docstring = bool(transformed_tree.body) and (
-                isinstance(transformed_tree.body[0], ast.Expr)
-                and isinstance(transformed_tree.body[0].value, AstStrType)
-                and (
-                    isinstance(transformed_tree.body[0].value.value, str)  # type: ignore[attr-defined]
-                    if sys.version_info >= (3, 8)
-                    else isinstance(transformed_tree.body[0].value.s, str)  # type: ignore[attr-defined]
-                )
-            )
-            transformed_tree.body.insert(1 if has_leading_docstring else 0, import_node)
+            # Insert after the module's mandatory preamble (docstring + ``from __future__``)
+            # so the injected import never displaces a statement Python requires to come first.
+            insert_index = self._preamble_insert_index(transformed_tree.body)
+            transformed_tree.body.insert(insert_index, import_node)
 
             # Compile transformed AST to code object
             code = compile(
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
@@ -1,0 +1,630 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+AST transformation for automatic function instrumentation.
+
+This module implements Python's import hook mechanism to automatically wrap
+all functions with ServiceEvents monitoring context managers during module loading.
+"""
+
+import ast
+import fnmatch
+import importlib
+import importlib.abc
+import importlib.util
+import logging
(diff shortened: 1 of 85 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
```diff
@@ -0,0 +1,397 @@
...
+                if exemplar and _endpoint_collector:
+                    _endpoint_collector.record_incident_exemplar(exemplar["operation"], exemplar)
+                # Mark incident as processed to prevent duplicate snapshots
+                g.serviceevents_incident_processed = True
+            # Telemetry must never crash the host app; snapshot processing is best-effort.
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                logger.error("Error processing incident snapshot: %s", exc, exc_info=True)
+    finally:
+        # Always clear operation context, even if there's an error
+        clear_current_operation()
```
jj22ee, line 304: **Investigation ContextVar is never cleared on the normal (non-incident) request path.**

`begin_investigation()` sets `_investigation_data` at the start of every request, but the only code that clears it is `get_investigation_data()` in `python_monitor_impl.py`, which is reached only *after* a snapshot is actually collected. For the common case (`trigger_type is None`), `process_potential_incident` returns early and the investigation dict is never cleared — the `finally` here only calls `clear_current_operation()`.

On pooled worker threads (Gunicorn sync, WSGI thread pools) the stale dict — including a live traceback object on error paths — lingers until the next request on that thread overwrites it via `begin_investigation()`. Cross-request bleed is bounded, but the frame retention on error paths is real.

Suggest adding an unconditional clear of investigation data in each framework's finalize/`finally` path (Flask teardown, FastAPI middleware `finally`, Django response path).

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
```diff
@@ -0,0 +1,645 @@
...
+            ast.fix_missing_locations(transformed_tree)
+
+            # Inject PythonServiceEventsMonitor import at the top of the module
+            import_node = ast.ImportFrom(
+                module="amazon.opentelemetry.distro.serviceevents.python_monitor",
+                names=[ast.alias(name="PythonServiceEventsMonitor", asname=None)],
+                level=0,
+            )
+            ast.fix_missing_locations(import_node)
+            # Insert the import after a leading module docstring (if any). A module's
```
jj22ee: **Inserting the monitor import at `body[0]` destroys the module-level docstring.**

`transformed_tree.body.insert(0, import_node)` places the injected `ImportFrom` before a leading module docstring expression, so `module.__doc__` becomes `None` for every instrumented module. (Function-level docstrings are preserved via `get_and_remove_docstring`, but the module-level one is not.)

This only affects packages a customer explicitly allowlists, but it's a silent behavior change for anything relying on `module.__doc__` (doctest, plugin loaders, CLIs that read `__doc__`). Suggest inserting *after* a leading docstring `Expr` if present (and after any `__future__` imports).

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
```diff
@@ -0,0 +1,397 @@
...
+    # Store original Flask.__init__ to wrap it
+    original_init = Flask.__init__
+
+    def instrumented_init(self, *args, **kwargs):
+        """Wrap Flask.__init__ to install hooks after app creation."""
+        # Call original init
+        original_init(self, *args, **kwargs)
+
+        # Install before_request and after_request hooks
+        self.before_request(_before_request_hook)
```
jj22ee: **`instrumented_init` is not crash-guarded — a failure here propagates out of the customer's `Flask(...)` constructor.**

The hook *bodies* (`_before_request_hook`, etc.) are guarded, but the wrapped `__init__` itself isn't. If `original_init` or any of the `before_request`/`after_request`/`teardown_request` registration calls raise, the exception escapes into the customer's app construction — telemetry breaking startup. Suggest wrapping the body of `instrumented_init` in a `try/except` that always calls `original_init` and swallows hook-registration failures.

(Double-install / idempotency is already handled by the `_initialized` guard in `serviceevents_instrumentation.py`, so this is just about crash-safety of the registration.)

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/fastapi_instrumentation.py
```diff
@@ -0,0 +1,451 @@
...
+    original_init = FastAPI.__init__
+
+    def instrumented_init(self, *args, **kwargs):
+        """Wrap FastAPI.__init__ to install middleware after app creation."""
+        # Call original init first and outside the guard below: this is the real
+        # FastAPI constructor and must always run. Only OUR added work is guarded,
+        # so a telemetry failure can never break FastAPI app construction.
+        original_init(self, *args, **kwargs)
+
+        # Crash-safety: telemetry must never break the host app. Installing the
```
jj22ee, line 443: **Same as Flask: `instrumented_init` isn't crash-guarded.**

If `original_init` or `add_middleware` raises (e.g. Starlette's `RuntimeError("Cannot add middleware after an application has started")` if init is ever re-entered), it propagates out of the customer's `FastAPI(...)` constructor. Suggest wrapping the body in `try/except` so a failure here can never break app construction.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/django_instrumentation.py
```diff
@@ -0,0 +1,445 @@
...
+            from django.conf import settings
+
+            middleware_list = list(getattr(settings, "MIDDLEWARE", []))
+
+            if _SERVICE_EVENTS_MIDDLEWARE_PATH not in middleware_list:
+                # Prepend to beginning (outermost middleware)
+                middleware_list.insert(0, _SERVICE_EVENTS_MIDDLEWARE_PATH)
+                settings.MIDDLEWARE = middleware_list
+
+                logger.info("ServiceEvents Django middleware prepended to settings.MIDDLEWARE")
```
jj22ee: **This mutates the global `settings.MIDDLEWARE` and runs the full `load_middleware` chain a second time.**

Two concerns:
1. Rewriting `settings.MIDDLEWARE` is a global side effect visible to customer code and other introspection; some apps validate/freeze settings.
2. Calling `original_load_middleware` again rebuilds the entire middleware chain on startup, re-running every customer middleware's `__init__` — so any middleware with import/init side effects runs twice.

It is crash-guarded and the `if path not in middleware_list` check prevents re-insertion, so it's not a blocker. But did you consider wrapping the handler / using a thin outer wrapper instead of mutating settings + double-loading? Also worth confirming there's no interaction with upstream OTel's own Django middleware injection.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
```diff
@@ -0,0 +1,397 @@
...
+    # Try to get the route pattern from the URL rule
+    if request.url_rule:
+        return request.url_rule.rule
+
+    # Fallback to endpoint name
+    if request.endpoint:
+        return f"/{request.endpoint}"
+
+    # Last resort: use path
+    return request.path
```
jj22ee: **Unmatched routes fall back to the raw path → metric cardinality risk.**

When there's no `url_rule`/`endpoint` (404s, unmatched URLs), this returns `request.path` verbatim. Scanner/bot traffic (`/.env`, `/wp-admin/...`, random URLs) each becomes a distinct `route` label, which can blow up endpoint-metric cardinality. The Django instrumentation already collapses unmatched requests to a `<unmatched>` sentinel — suggest the same here (and in FastAPI) for parity.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/fastapi_instrumentation.py
```diff
@@ -0,0 +1,451 @@
...
+    if route and hasattr(route, "path"):
+        return route.path  # e.g., "/api/users/{id}"
+
+    # Not yet routed: try to pre-resolve the template against the app's routes.
+    resolved = _resolve_route_template(scope)
+    if resolved is not None:
+        return resolved
+
+    # Fallback to path
+    return scope.get("path", "/unknown")
```
jj22ee: **Same cardinality concern as Flask:** unmatched routes fall back to `scope["path"]`, so 404/scanner traffic produces a distinct `route` label per URL. Suggest a `<unmatched>` sentinel to match Django's behavior.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/config.py
```diff
@@ -0,0 +1,518 @@
...
+
+    return None
+
+
+# Internal test-config hook. NOT a customer-facing surface — undocumented, gated, and applied
+# only when DEBUG_SE_TEST_CONFIG is set. Black-box contract/e2e suites run the SDK in a separate
+# process and can only inject configuration via env; this lets them override the small set of
+# internal knobs that no longer have their own env vars. Format is dependency-free delimited
+# "KEY=value;KEY=value" where KEY is the former env-var suffix.
+_TEST_CONFIG_HOOK_ENV = "DEBUG_SE_TEST_CONFIG"
```
jj22ee, line 83: **Should `DEBUG_SE_TEST_CONFIG` ship in the public distro, or be stripped for prod?**

This is a production-reachable hook that lets an env var override internal config (`log_group`, `log_stream`, flush intervals, sampling tiers) that otherwise have no env var. It's reasonably safe as written — gated (no-op when unset), warn-once, a fixed allowlist of fields, and unknown/unparsable keys ignored — and I understand the rationale (black-box contract/e2e suites run the SDK in a separate process and can only inject config via env).

But it's still an unsupported config backdoor shipping in the public package. Is the intent to keep it in production, or should it be gated behind a non-prod build / test-only path? Not blocking — just want a conscious decision on whether it belongs in the released artifact.

jj22ee (commented):
Added some comments generated through Claude based on AI review.

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/ast_transformation.py
vastin: fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/django_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/fastapi_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/flask_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/fastapi_instrumentation.py
vastin: Fixed

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/config.py
vastin: Keep since contract test can only inject config via env and it's gated, warn-once, allowlisted, and a no-op when unset.

jj22ee (dismissed):
Apart from the bot's concern regarding unintended suppression of original application exception, lgtm from my side

jj22ee (approved):

mxiamxia (approved):

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/serviceevents/instrumentation/fastapi_instrumentation.py
```diff
@@ -0,0 +1,459 @@
...
+                )
+                await send(
+                    {
+                        "type": "http.response.body",
+                        "body": b"Internal Server Error",
+                    }
+                )
+
+            # Re-raise to maintain error propagation
+            raise
```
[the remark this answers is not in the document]
vastin: fixed

PR #766, opened 2026-06-11
fix(debugger): populate service.name on DI snapshot OTLP exports

liustve:
## Bug

`SnapshotOtlpEmitter` (added in #761) builds its default `LoggerProvider` with `Resource.get_empty()`, which skips the OTel SDK detector chain. `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` are never read, so DI snapshot LogRecords ship with no `service.name`. The CW Agent transform then routes them to `/aws/service-events/unknown_service` instead of the customer's log group.

Other DI paths (status reporter, snapshot body) read `OTEL_SERVICE_NAME` directly via `os.environ.get(...)`, so only the OTLP resource path was broken.

## Fix

[_snapshot_otlp_emitter.py:79](aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_snapshot_otlp_emitter.py#L79)

```diff
-resource = self._resource if self._resource else Resource.get_empty()
+resource = self._resource if self._resource else Resource.create()
```

`Resource.create()` runs the standard detector chain (`OTELResourceDetector`, ...) which reads `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES`. Explicit resources passed by callers are still preserved.

## Tests

Three regression tests in `test_snapshot_otlp_emitter_extra.py`, all verified to fail on `Resource.get_empty()` and pass on `Resource.create()`:
- `OTEL_SERVICE_NAME` lands on the resource
- `OTEL_RESOURCE_ATTRIBUTES` keys merge in alongside it
- `OTEL_SERVICE_NAME` wins over `service.name=...` inside `OTEL_RESOURCE_ATTRIBUTES`

Plus one test confirming an explicit `Resource` passed to the constructor is preserved unchanged.

## Validated end-to-end

EC2 soak with `OTEL_SERVICE_NAME=python-ec2-di-main-service`: snapshots route to `/aws/service-events/python-ec2-di-main-service` and are queryable by `aws.di.location_hash`.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_snapshot_otlp_emitter.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_snapshot_otlp_emitter.py
@@ -76,7 +76,7 @@ class SnapshotOtlpEmitter:
 
             try:
                 exporter = OTLPLogExporter(endpoint=self._logs_endpoint)
-                resource = self._resource if self._resource else Resource.get_empty()
+                resource = self._resource if self._resource else Resource.create()
 
                 self._logger_provider = LoggerProvider(resource=resource)
                 self._logger_provider.add_log_record_processor(BatchLogRecordProcessor(exporter))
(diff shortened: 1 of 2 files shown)


vastin (approved):
LGTM

PR #772, opened 2026-06-15
fix(debugger): use bytecode/sys.monitoring instead of setattr wrapper

liustve:
## Problem

The current Dynamic Instrumentation path replaces a target function with a Python-level wrapper installed via `setattr(module, name, wrapper)`. This has two issues, one architectural and one downstream of it.

**1. setattr only updates the module's name binding.** Frameworks register handler functions at import time and store direct references in their own internal data structures (route registries, dispatcher tables, ORM hooks, etc.). When a request comes in, the framework calls those stored references, not `module.foo`. So setattr-replacing `module.foo` leaves DI silently bypassed for every framework-registered entry point — the wrapper never runs, no snapshot, no instrumentation. The previous workaround was a per-framework patcher that scanned for known frameworks and rewrote their internal dicts. This is fundamentally a losing approach: every new framework needs its own patcher, the matching heuristics break under decorator wrappers, and any framework we don't know about is invisible.

**2. The wrapper's exception path drops the throwable.** When a wrapped function raises, the resulting snapshot's `Captures.return_context.throwable` field is empty — no captured exception type, message, or stacktrace. The user gets a snapshot that says the function failed but provides none of the data they need to diagnose it. The cause is that the wrapper's `try/except` sits on top of the user's frame, and exception-state lifecycle differences between Python versions (especially 3.11+ with PEP 657) made the capture logic unreliable.

Both problems share a root cause: instrumenting at the Python layer instead of at the interpreter layer.

## Fix

Instrument at the interpreter layer. The function's `__code__` is mutated in place — any reference any framework or library is holding sees the rewritten code on the next call, with no per-framework patching. The interpreter's own exception flow runs the unwind hook, so `sys.exc_info()` is reliably the live exception and the throwable is always captured.

Each Python version uses the engine that fits it best, and all three engines share the same handler contract: entry, return, unwind.

| Python | Engine | Mechanism |
|---|---|---|
| 3.12+ | SysMonitoringEngine | PEP 669 (`sys.monitoring`): PY_START / PY_RETURN / PY_UNWIND |
| 3.9 / 3.10 | BytecodeInjectionEngine | `SETUP_FINALLY`-based body wrap, RERAISE in handler |
| 3.11 | BytecodeInjectionEngine | PEP 657 `co_exceptiontable` wrap via splice walk (see below) |

The unwind handler reads `sys.exc_info()` directly, builds a `CapturedThrowable` with type/message/stacktrace (in-function frames merged with caller frames), and emits the snapshot.

### 3.12+: sys.monitoring
Native PEP 669 events. Per-thread LIFO stack of `(code_id, start_ns)` tuples preserves correct durations across recursion. PY_UNWIND is set globally (it isn't local-event-capable on 3.12-3.14); a cheap `_function_keys` membership check filters to instrumented codes before any allocation. Reentrancy guard via `threading.local` so a snapshot built from the serializer cannot itself trigger another snapshot on the same thread.

### 3.9 / 3.10: bytecode rewrite with SETUP_FINALLY
Standard `SETUP_FINALLY` body wrap: prologue calls the entry hook, every `RETURN_VALUE` is replaced with a sequence that calls the exit hook then returns, and the SETUP_FINALLY block calls the unwind hook then `RERAISE`s. Local slots for `start_ns` / `entry_ctx` / `retval` are pre-initialized to `None` so the unwind hook's `LOAD_FAST` is safe even if the entry hook crashed before storing them. `POP_BLOCK` is emitted before each `RETURN_VALUE` so the block stack is balanced on the normal path.

### 3.11: bytecode rewrite with PEP 657 exception_table — splice walk
3.11 removed `SETUP_FINALLY`. Protected regions live in out-of-band metadata (`co_exceptiontable`). The bytecode library exposes this via `TryBegin` / `TryEnd` pseudo-instructions, but it forbids nested `TryBegin` — and any function that contains user `try/except` already has one. Wrapping the whole body with another `TryBegin` raises `RuntimeError: TryBegin pseudo instructions cannot be nested`.

The fix is a splice walk that produces N adjacent (never nested) outer regions, all targeting one shared exception handler. Walking the user's instructions:

```
            (preamble: RESUME, COPY_FREE_VARS, MAKE_CELL)
            <prologue: call entry hook>
[outer]     TryBegin(handler_label, push_lasti=True)
              ... body before user try ...
[outer]     TryEnd                          # close BEFORE user try
              user_try_begin
                ... user-protected body ...
              user_try_end
[outer]     TryBegin(handler_label, ...)    # reopen AFTER user try
              ... body after user try ...
              <each RETURN_VALUE replaced with: exit hook + return>
[outer]     TryEnd
            handler_label:
              <unwind hook + RERAISE>
```

The outer protected regions are flat (one closes before each user try, the next opens after) so the library accepts them, and they all funnel uncaught exceptions to the same handler. Other 3.11-specific details:

- Prologue is inserted **after** `RESUME 0` / `COPY_FREE_VARS` / `MAKE_CELL` so closures and free-vars are armed when the entry hook fires (otherwise `frame.f_locals` is incomplete).
- 3.11 calling convention: `PUSH_NULL` + `LOAD_GLOBAL((push_null: bool, name: str))` + `PRECALL n` + `CALL n` (the tuple-form `LOAD_GLOBAL` arg is mandatory; passing a bare string silently miswrites `namei`).
- Handler entry stack with `push_lasti=True` is `[..., lasti, exc]`. 3.11 does NOT auto-set `sys.exc_info()` for `co_exceptiontable` handlers, so the foot uses `PUSH_EXC_INFO` to expose the live exception to the unwind hook, then `RERAISE 0` to propagate.

### Decorator unwrapping
`@functools.wraps`-style decorators rebind `module.name` to a wrapper whose `__code__` is the decorator's check, not the user's body. Without unwrapping, the engine would rewrite the decorator's code and snapshots would report the wrapper's location and fire on every decorator invocation instead of the actual handler. A new `_undecorate.undecorated()` BFSes through `__wrapped__` / `functools.partial` / closure cells, with `co_filename` disambiguation, to find the real user function. The engine arms that one and keys state by its id.

### Async / generator fallback
Engines decline to rewrite generators, coroutines, async generators, and iterable coroutines (rewriting their bodies would corrupt `.send()` / `.throw()` / `await` semantics). On decline, the manager routes through `FunctionWrapper.instrument_function`, which has dedicated coroutine support via `_create_async_wrapper`. So `async def` functions remain instrumented end-to-end — just not by the bytecode engine.

### Framework-specific patcher removed
The `__code__` mutation is in-place: every reference any framework holds sees the rewritten code on the next call. No framework-specific scanning, no per-framework patcher, no fragile name+module identity matching. The previously-needed patcher was deleted and its tests rewritten in place to exercise the engine path and prove the framework's stored reference flows through without explicit patching.

### State leak fixes
- `_call_start_ns` is popped unconditionally in PY_RETURN/PY_UNWIND before the reentrancy/membership guards (otherwise a disable-mid-flight or serializer-triggered short-circuit orphans the stamp).
- `disable_function_level_instrumentation` clears all per-code state including `_instrumentation_types` and the per-thread stack.
- `disable_breakpoints_for_function` (the existing line-BP path) also clears function-level state on the same code, so coexisting instrumentation tears down cleanly.

## Other changes in this PR

- `SnapshotOtlpEmitter.initialize()` is called eagerly during manager init. Lazy init from a `sys.monitoring` callback at shutdown could hit `RuntimeError(\"cannot schedule new futures after interpreter shutdown\")` because `BatchLogRecordProcessor` spawns a daemon worker on construction.
- `_build_resource` writes both `deployment.environment` (older convention) and `deployment.environment.name` (OTel-stable) so consumers reading either key see the value. Engines' `_get_environment` helpers prefer the `.name` variant and fall back to the bare key.
- All hardcoded `\"service.name\"` / `\"deployment.environment(.name)\"` strings replaced with constants from `opentelemetry.semconv.resource.ResourceAttributes` and `opentelemetry.semconv._incubating.attributes.deployment_attributes.DEPLOYMENT_ENVIRONMENT_NAME`. All inline OTel imports hoisted to module top.
- Conftest pins OTel semconv schema URL to `v1.41.0` raw GitHub path so schema-validation tests are stable when the docs site drifts.

## Validation

| Python | Tests passing | Function-level scenarios |
|---|---|---|
| 3.9 | 528 | basic, throwable, generator/coroutine/async-gen decline, disable-restores, kwonly+varargs, recursion, throwable-inside-user-try, decorator unwrap, disable-after-decorated-enable |
| 3.10 | 528 | same as 3.9 |
| 3.11 | 528 | same as 3.9 (10 tests previously skipped on 3.11 are now active and passing) |
| 3.12 | 523 | + 2 SME recursion-duration regression tests |

- \`black\` / \`isort\` / \`flake8\`: clean across 47 files.
- Regression harness against \`origin/main\` (12 scenarios): PASS, 0 unexpected deltas.
- E2E test framework: line-level + function-level + exception-path snapshots validated in CW Logs (sys.monitoring path on 3.12 and bytecode path on 3.9/3.10/3.11 all green).

## Test plan

- [x] Unit tests pass on 3.9 / 3.10 / 3.11 / 3.12
- [x] Linters clean
- [x] Regression harness shows zero deltas vs origin/main
- [x] E2E test framework validates throwable capture end-to-end


liustve:
Keeping it as draft to give more time to evaluate.

PR #786, opened 2026-06-18
fix(dynamic-instrumentation): redirect `from module import func` aliases for method-level breakpoints

srprash:
## Description

A method-level (function-level) Dynamic Instrumentation breakpoint installs its wrapper by `setattr`-ing the wrapper onto the **defining module's** attribute (plus patching Flask/Django/FastAPI route tables). When a caller imports the target with `from module import func` and then calls the **bare name**, that caller holds an *independent* reference in its own module namespace. `setattr` on the defining module does not update that copy, so the caller keeps invoking the original, un-instrumented function.

The result is a silent failure: the breakpoint reports **READY**, no error is logged, but it **never fires** and no snapshot is produced. `from module import func` is a common, PEP 8-sanctioned import style, so this affects a large class of real applications. (Line-level breakpoints are unaffected because the `sys.monitoring` engine binds the code object directly.)

## Fix

`FunctionWrapper` now redirects those module-level aliases on install (and restores them on uninstall) via a new `_patch_module_aliases` helper, generalizing the existing framework-reference patching to plain module-level references.

The scan is deliberately bounded for safety (this is a safety-critical agent — it must never break the user's application):

- **Application modules only.** `_is_application_module` excludes the standard library, `site-packages`/`dist-packages` (third-party deps), built-in/frozen/namespace modules, and the OpenTelemetry/ADOT SDK itself. Classification fails closed (when in doubt, don't touch).
- **Identity match only** (`attr is original_func`) — never name-based guessing.
- **Module-level attributes only** — never reaches into dicts/registries/containers.
- **`sys.modules` and each module's `vars()` are snapshotted** before iterating, so a concurrent import cannot cause "dictionary changed size during iteration."
- **Every `setattr` is individually guarded**, and the helper never raises — a failure to rewrite one reference never aborts the rest or the poller.
- **Symmetric on uninstall** — deleting the breakpoint restores the aliases to the original, returning the app to its prior state.

### Performance

The scan runs **once per method-level breakpoint install/uninstall, on the polling thread** — not on the application request path, and not per poll (gated by the manager's change detection). The per-call wrapper is unchanged, so steady-state cost of an active breakpoint is unaffected. `realpath()` of the stdlib prefixes is computed once per scan (not per module), and module classification is memoized per scan.

## Scope / known limitations

- References captured into **local variables, closures, default arguments, or data structures** cannot be reached by any name-based rebind; those remain out of scope (use a line-level breakpoint).
- Applications installed/run from **`site-packages`** (e.g. `pip install .`, or copied into the image's site-packages) classify as non-application and are not rewritten — fail-safe, but a coverage gap; a follow-up could key "application module" off the defining module's own package tree.
- This change is independent of and complementary to the FastAPI `Depends()` sub-dependency rebind (#785) — different reference location (route dependant tree vs. module namespace), no overlap.

## Testing

- **Unit tests** (`test_function_wrapper.py::TestModuleAliasRedirect`): alias redirected on install; defining module skipped; stdlib and site-packages modules untouched; restore reverts; an un-writable namespace does not raise. Full debugger unit suite: 550 passed.
- **Contract test** (`flask_test.py::DIFlaskFromImportAliasTest`): a new `pricing_service.apply_discount` target reached via `from pricing_service import apply_discount` in `di_flask_server`; asserts a method-level snapshot with the correct arguments and return value is produced through the alias. Verified as a genuine regression guard — it passes with the fix and fails without it (the breakpoint installs but never fires).
- Lint clean (black, isort, flake8) and pylint 10.00/10.

### DIFF
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_function_wrapper.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_function_wrapper.py
@@ -244,6 +244,9 @@ class FunctionWrapper:
                 # Restore framework-level references back to the original function
                 if wrapped_func is not None:
                     FunctionWrapper._patch_framework_references(module, wrapped_func, original_func)
+                    # Symmetrically restore any `from <module> import <func>` aliases we redirected
+                    # at install time, so deleting the breakpoint returns the app to its prior state.
+                    FunctionWrapper._patch_module_aliases(module, wrapped_func, original_func)
 
             return True
         except Exception as exception:
@@ -776,6 +779,17 @@ class FunctionWrapper:
                 # unwrapped function, bypassing DI instrumentation entirely.
                 if original_func is not None:
                     FunctionWrapper._patch_framework_references(module, original_func, new_func)
+                    # Also redirect plain `from <module> import <func>` aliases held in other
+                    # application modules (the bare-name call site keeps a reference to the
+                    # original function, which setattr on the defining module does not update).
+                    alias_count = FunctionWrapper._patch_module_aliases(module, original_func, new_func)
+                    if alias_count:
+                        logger.debug(
+                            "DI: redirected %d module-level alias(es) of %s.%s",
+                            alias_count,
+                            module_name,
+                            function_name,
+                        )
 
         except ImportError as exception:
             logger.error("Failed to import module '%s' for replacement: %s", module_name, exception)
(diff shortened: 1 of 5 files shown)


aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_function_wrapper.py
```diff
@@ -823,6 +837,151 @@ def _patch_framework_references(module, original_func: Callable, new_func: Calla
...
+            spec = getattr(module, "__spec__", None)
+            origin = getattr(spec, "origin", None) if spec is not None else getattr(module, "__file__", None)
+            # Built-in / frozen / namespace modules have no real file origin — skip them.
+            if not origin or origin in ("built-in", "frozen", "namespace"):
+                return False
+            origin = os.path.realpath(origin)
+
+            # Skip anything under installed packages.
+            for marker in ("site-packages", "dist-packages"):
+                if marker in origin:
```
liustve: `marker in origin` is a substring test, not a path-component check, so it produces a silent false-negative: any legitimate application path that merely *contains* the text `site-packages` / `dist-packages` (e.g. a project at `/home/dev/site-packages-demo/myapp/server.py`, or a dir a user named `my-site-packages`) classifies as non-application and its `from x import f` aliases are never redirected. The breakpoint still installs and reports READY, but never fires for that app — the exact silent failure this PR is fixing, re-entering through a path-naming edge case.

Reproduced against this branch with the real `_patch_module_aliases`:
```
use.__file__ : /tmp/.../site-packages-demo/myapp/use.py
is_application_module(use): False
rebound count: 0
use.func after install: ORIGINAL (silently missed!)
```

Suggest matching a real path component instead of a substring:
```python
parts = origin.split(os.sep)
if "site-packages" in parts or "dist-packages" in parts:
    return False
```

(The adjacent `origin.startswith(libdir)` stdlib check on the next lines has the same flavor — an app deployed under `<venv>/lib/...` misclassifies as stdlib — and when `__spec__` exists, `__file__` is never used as a fallback. Flagging here since they're the same root cause; happy to file separately.)

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_function_wrapper.py
```diff
@@ -823,6 +837,151 @@ def _patch_framework_references(module, original_func: Callable, new_func: Calla
...
+                    return False
+            # Skip the standard library (under sys.prefix/lib but not the app's own tree).
+            for libdir in stdlib_lib_dirs or FunctionWrapper._stdlib_lib_dirs():
+                if origin.startswith(libdir):
```
liustve, line 925: Filing the second half of the classification cluster I mentioned on line 899. Two fail-safe-but-silent false-negatives here, both dropping alias coverage for legitimate first-party code:

**1. `origin.startswith(libdir)` misclassifies apps deployed under the prefix.** `libdir` is `<sys.prefix>/lib` / `<sys.base_prefix>/lib`. Some real deploy layouts place app code under the environment prefix — e.g. PyInstaller/embedded layouts, or images that copy the app into `<venv>/lib/...`. Such an app's path starts with `libdir`, so it classifies as stdlib → `False`, and its `from x import f` aliases are never redirected (silent READY, never fires). The PR's stated limitation only calls out apps installed *under* `site-packages`; this `<prefix>/lib` case isn't mentioned.

Reproduced on this branch — an app module whose origin is built under `_stdlib_lib_dirs()[0]`:
```
is_application_module(under <prefix>/lib): False   # wanted True (legit app)
```

**2. `__file__` is never used as a fallback (line 891).** `origin` is taken from `spec.origin` whenever `__spec__` exists, otherwise `__file__`. But if `__spec__` exists with a `None`/non-path `origin` (some loaders, namespace-ish or dynamically-created modules) while `__file__` is a perfectly good app path, the module is dropped at the `if not origin` guard on line 893 even though `__file__` would have classified it as app.

Both are safe (no crash, no wrong rewrite) but invisibly reduce coverage. A path-component check (as suggested on line 899) plus falling back to `__file__` when `spec.origin` is unusable would cover all four cases in this cluster:
```python
origin = (getattr(spec, "origin", None) if spec is not None else None) or getattr(module, "__file__", None)
...
parts = os.path.realpath(origin).split(os.sep)
if "site-packages" in parts or "dist-packages" in parts:
    return False
```

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/debugger/_function_wrapper.py
```diff
@@ -823,6 +837,151 @@ def _patch_framework_references(module, original_func: Callable, new_func: Calla
...
+
+        Note: apps installed under ``site-packages``/``dist-packages`` (e.g. ``pip install .`` or
+        copied into the image's site-packages) classify as non-application and are NOT rewritten,
+        so the alias redirect does not reach them. This is fail-safe (no rewrite, no crash) but is
+        a known coverage limitation.
+        """
+        try:
+            name = getattr(module, "__name__", "") or ""
+            for prefix in FunctionWrapper._ALIAS_SKIP_PREFIXES:
+                if name == prefix or name.startswith(prefix):
```
liustve: Same loose-matching root cause as the two `origin` checks below (substring / `startswith` where a structural match was needed), but on module *name* this time. Three of the five entries — `wrapt`, `_pytest`, `pytest` — have no trailing dot, and they're matched with `name.startswith(prefix)`. So any first-party app module whose name merely *begins* with those letters is silently treated as non-application and its `from x import f` aliases are never redirected (installs, reports READY, never fires).

Reproduced on this branch:
```
is_application_module('wraptools')           = False   # wanted True (legit app)
is_application_module('wrapt_helpers')        = False
is_application_module('pytest_fixtures_local')= False
```
(`amazon.opentelemetry.` and `opentelemetry.` are fine — they have the trailing dot.)

Suggest matching the first dotted component for top-level names, keeping `startswith` only for the dotted package trees:
```python
top = name.split(".", 1)[0]
if top in ("wrapt", "_pytest", "pytest") or name.startswith(("amazon.opentelemetry.", "opentelemetry.")):
    return False
```
This still skips `wrapt`, `wrapt.decorators`, `pytest`, `pytest.fixtures`, etc., but no longer eats `wraptools` / `pytest_fixtures`.

liustve (approved):

wangzlei (approved):

PR #788, opened 2026-06-18
ci: parallelize and cache contract-test builds

liustve:
Speeds up the ~1hr PR build by restructuring the contract-test pipeline:

- **Shard contract tests** by suite × Python version into parallel jobs (one collapsible group per version) instead of a single serial run.
- **Build the distro wheel once** and reuse it across unit tests and every contract-image build.
- **Cache Docker layers** (`type=gha`) so the operator-image Rust stage and dependency installs persist across runs.
- **Parameterize app images** with `PYTHON_VERSION` so each version is genuinely exercised; bump stale deps (django, psycopg2, mysql-connector, typing-extensions) to extend coverage to 3.13/3.14.
- DI sharded by framework; lambda tests run the full 3.10–3.14 matrix.


scripts/set-up-contract-tests.sh
```diff
@@ -13,12 +13,33 @@ if [ "$current_dir" != "aws-otel-python-instrumentation" ]; then
   exit
 fi
 
+PYTHON_VERSION="${1:-}"
```
ezhang6811, line 16: Can you update the contract test README if this script requires additional args now?

ezhang6811:
Thanks for this change, glad to see improvements in our CI efficiency.

scripts/set-up-contract-tests.sh
liustve: These args are optional

ezhang6811 (approved):

wangzlei (approved):

PR #789, opened 2026-06-18
feat: add OTel lite SDK for Lambda cold start optimization

wespil:
# otel-lite-sdk

A lightweight OpenTelemetry SDK replacement optimized for AWS Lambda cold starts. It replaces the full opentelemetry-sdk tracing pipeline with a minimal implementation that provides the same OTel API contract.

### Core tracing pipeline (opentelemetry_lite_sdk.py):

- **TracerProvider** — extends trace_api.TracerProvider, holds a list of span processors and dispatches on_start/on_end directly
- **Tracer** — extends trace_api.Tracer, implements start_span and start_as_current_span with parent-based always-on sampling
- **Span** — lock-free span implementation (Lambda is single-threaded), supports set_attribute, add_event, set_status, record_exception, update_name
- **InstrumentationScope** — scope container with name, version, schema_url, and attributes (fully stored, grouped, and encoded)
- **Event** — lightweight span event with name, attributes, and timestamp
- **SpanProcessor** — base class with on_start/on_end/force_flush/shutdown interface
- **BatchingSpanProcessor** — collects spans during invocation and exports all on force_flush (no background threads)
- **UdpSpanExporter** — encodes spans to OTLP protobuf and sends via UDP to the X-Ray daemon; injects Application Signals attributes when enabled
- **UdpExporter** — non-blocking UDP socket transport with X-Ray daemon protocol framing (T1S/T1U prefix + base64)
- **OTLP Protobuf Encoder** — hand-rolled encoder for ExportTraceServiceRequest (no protobuf library dependency); supports all AnyValue types including arrays
- **Resource** — built from OTEL_RESOURCE_ATTRIBUTES and OTEL_SERVICE_NAME env vars at init time (stored as dict)
- **Parent-Based Always-On Sampling** — root spans are always sampled; child spans inherit `trace_flags` from the propagated parent context
- **Span Flags** — encodes HAS_IS_REMOTE (0x100) and IS_REMOTE (0x200) bits for proper backend interpretation
- By default it manually instruments `Botocore`, `requests`, `urllib3`

## Activating Fast Start Mode

Add a new environment variable: `OTEL_AWS_LAMBDA_FAST_START = true`

## What's supported vs. the standard SDK path

| Feature / Env Var | Standard SDK | Fast Start (Lite) | Behavior in Fast Start |
|---|---|---|---|
| `OTEL_RESOURCE_ATTRIBUTES` | Honored | Honored | Parsed at init time |
| `OTEL_SERVICE_NAME` | Honored | Honored | Overrides `service.name` from resource attrs only when non-empty |
| `AWS_XRAY_DAEMON_ADDRESS` | Not used | Honored | UDP endpoint for trace export |
| `OTEL_AWS_APPLICATION_SIGNALS_ENABLED` | Honored | Honored | Injects aws.local/remote attributes |
| `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` | Honored | Honored | Subtracts from the hardcoded enabled list |
| `OTEL_PROPAGATORS` | Honored | Silently ignored | Hardcoded: xray + tracecontext + baggage |
| `OTEL_TRACES_SAMPLER` | Honored | Silently ignored | Always samples root spans |
| `OTEL_TRACES_EXPORTER` | Honored | Silently ignored | Always UDP to X-Ray daemon |
| `OTEL_EXPORTER_OTLP_ENDPOINT` | Honored | Silently ignored | Uses `AWS_XRAY_DAEMON_ADDRESS` |
| `OTEL_METRICS_EXPORTER` | Honored | No metrics support | Silently ignored |
| `OTEL_LOGS_EXPORTER` | Honored | No logs support | Silently ignored |
| Custom SpanProcessor | Supported | Not supported | Only built-in BatchingSpanProcessor |
| Custom Exporter | Supported | Not supported | Only UDP to X-Ray daemon |
| Span Links | Encoded | Accepted but dropped | Parameter kept for API compatibility |
| InstrumentationScope attributes | Stored | Fully supported | Stored, grouped by, and encoded in OTLP |

### Bootstrap (otel_wrapper.py)

- Detects `OTEL_AWS_LAMBDA_FAST_START=true`, calls `configure_lite_mode()` which:
  - Instantiates TracerProvider + BatchingSpanProcessor + UdpSpanExporter
  - Registers as the global TracerProvider via `trace_api.set_tracer_provider()`
  - Sets up composite propagator (Baggage + X-Ray + W3C TraceContext)
  - Instruments botocore, requests, and urllib3
- Logs errors with full traceback if `configure_lite_mode()` fails (no silent swallowing)
- Then `AwsLambdaInstrumentor().instrument()` runs using the lite TracerProvider

### Lambda wrapper (otel-instrument)

- Defaults `OTEL_AWS_LAMBDA_FAST_START` to `"false"` if not set
- When `OTEL_AWS_LAMBDA_FAST_START=true`: skips `opentelemetry-instrument` auto-instrumentation bootstrap, passes through directly to the runtime (`exec "$@"`); disabled list default is not set (enabled list in `configure_lite_mode()` is authoritative)
- When `OTEL_AWS_LAMBDA_FAST_START=false`: falls back to the standard `opentelemetry-instrument` path with the full disabled instrumentations list

### Application Signals support

- When `OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true`, the exporter injects `aws.local.service`, `aws.local.operation`, `aws.local.environment`, `aws.remote.service`, and `aws.remote.operation` attributes onto spans before OTLP encoding
- Spans are sent via UDP with T1S (sampled) or T1U (unsampled) prefix based on the parent's sampling decision
- Service operations and dependencies appear in Application Signals without requiring EMF metrics or the ADOT collector

### Code correlation gating (aws_lambda/__init__.py)

- When `OTEL_AWS_LAMBDA_FAST_START=true`, skips importing `code_correlation` module (avoids pulling in opentelemetry.sdk.trace)
- Falls back to a no-op `add_code_attributes_to_span` function


aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/test_opentelemetry_lite_sdk.py
```diff
@@ -0,0 +1,540 @@
...
+        span.set_attribute("rpc.system", "aws-api")
+        span.set_attribute("rpc.method", "ListBuckets")
+        span.end()
+
+        exporter.export([span])
+        self.assertEqual(span._attributes["aws.remote.service"], "AWS::S3")
+        self.assertEqual(span._attributes["aws.remote.operation"], "ListBuckets")
+
+    @patch.dict(os.environ, {"OTEL_AWS_APPLICATION_SIGNALS_ENABLED": "false"})
+    def test_export_uses_t1s_prefix_for_sampled(self):
```
ezhang6811, line 419: add an equivalent unsampled - T1U test

lambda-layer/src/otel-instrument
```diff
@@ -128,7 +132,15 @@ else
     export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
 fi
 
-# - Enable botocore instrumentation by default
+# - Set the enabled instrumentations allowlist for lite mode.
+#   If OTEL_PYTHON_ENABLED_INSTRUMENTATIONS is set, only those libraries are instrumented.
+#   This is read by configure_lite_mode() in the lite SDK.
+
+if [ -z "${OTEL_PYTHON_ENABLED_INSTRUMENTATIONS}" ] && [ -z "${OTEL_PYTHON_DISABLED_INSTRUMENTATIONS}" ]; then
+    export OTEL_PYTHON_ENABLED_INSTRUMENTATIONS="botocore,requests,urllib3"
```
wangzlei: OTEL_PYTHON_ENABLED_INSTRUMENTATIONS is not available in otel python https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1828

Do we want to create a new environment variable specifically for the new mode?

lambda-layer/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -91,14 +91,20 @@ def custom_event_context_extractor(lambda_event):
 from opentelemetry.trace import Span, SpanKind, TracerProvider, get_tracer, get_tracer_provider
 from opentelemetry.trace.status import Status, StatusCode
 
-# Import code correlation functionality
-try:
-    from amazon.opentelemetry.distro.code_correlation import add_code_attributes_to_span
-except ImportError:
-    # If code correlation module is not available, define no-op functions
+# Import code correlation functionality (skip in lite mode to avoid SDK imports)
+if os.environ.get("AWS_LAMBDA_LITE_MODE", "false").lower() == "true":
```
wangzlei: "lite" describes how it's built, should we name this after the benefit rather than the implementation? e.g. OTEL_AWS_LAMBDA_FAST_START=true ?

aws-opentelemetry-distro/src/amazon/opentelemetry/distro/opentelemetry_lite_sdk.py
```diff
@@ -0,0 +1,802 @@
...
+        CompositePropagator(
+            [
+                W3CBaggagePropagator(),
+                AwsXRayPropagator(),
+                TraceContextTextMapPropagator(),
+            ]
+        )
+    )
+
+    enabled = set(d.strip() for d in os.environ.get("OTEL_PYTHON_ENABLED_INSTRUMENTATIONS", "").split(",") if d.strip())
```
wangzlei: Better to not introduce new environment variable

wangzlei (changes requested):

wangzlei:
Could we add a section listing what this mode supports vs. doesn't support compared to the standard SDK path? For example, what happens if a user tries to override the propagator, sampler, span processor, exporter, or resource? Are these env vars (e.g. OTEL_PROPAGATORS, OTEL_TRACES_SAMPLER) silently ignored, partially honored, or do they cause an error?

wangzlei (approved):

PR #800, opened 2026-06-25
ci: parallelize main-build & refactor build/test actions

liustve:


liustve:
Split into two stacked draft PRs: #801 (build/test action refactor, base main) and #802 (main-build parallelization, stacked on #801).

PR #817, opened 2026-07-09
fix(genai): match OTel util-genai tool value and bytes serialization

liustve:
Aligns tool call argument/result serialization with OTel util-genai's `_any_value_to_attribute_value`: primitives (bool/str/bytes/int/float) are kept as native span attribute values instead of being JSON-quoted, while dicts/lists are JSON-serialized. Also fixes `serialize_to_json_string` to base64-encode bytes (matching util-genai's JSON encoder) instead of silently dropping them, so multimodal blob content survives. Applied via a shared `to_tool_attribute_value` helper across crewai, langchain, llama_index, and mcp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- fix(genai): serialize tool call arguments/results and blob bytes to match OTel util-genai (primitives kept native, bytes base64-encoded)
+  ([#817](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/817))
 - feat(genai): capture user input and agent output on invoke_agent spans
   ([#815](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/815))
 - refactor(serviceevents): make the endpoint span processor framework-agnostic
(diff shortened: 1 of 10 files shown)


liustve:
Thanks for the review! Addressing the points:

1. **Blob size** — acknowledged & intentional (util-genai parity). Noted as the first place to look if exporter payloads grow.
2. **Docstring URL** — the docstring (and the `_GenAIJSONEncoder` subclass it described) was already removed in a follow-up refactor; `serialize_to_json_string` now uses `json.dumps(..., default=lambda o: b64encode(o).decode())`, so there's no dangling reference.
3. **Combined dict-with-bytes test** — added `test_to_tool_attribute_value_dict_with_bytes_base64_encoded` locking the blob-survival contract end-to-end (a8f4c0a).
4. **`_set_span_attribute` filter** — confirmed `False`/`0` still set correctly (type-unequal to `""`). One edge case worth noting: an **empty-string** tool result now flows through as `""` (dropped by the pre-existing `value != ""` filter) instead of the old `'""'` (kept). This is langchain-specific, pre-existing filter behavior and an empty result carries no content, so I left it as-is.

wangzlei (approved):

PR #818, opened 2026-07-16
refactor(serviceevents): remove direct-to-CloudWatch SigV4 OTLP log export path

liustve:
Removes the direct-to-CloudWatch SigV4 OTLP log export path from ServiceEvents. This path was originally added for testing and is no longer needed — ServiceEvents exports through the collector-proxied OTLP endpoint.


vastin (approved):
LGTM

PR #827, opened 2026-07-17
fix(mcp-instrumentation): inject W3C trace context into outbound HTTP headers

### LINKED ISSUE #826: MCP instrumentation breaks trace propagation to MCP servers that read context from HTTP headers (defaults hide the problem)
# Summary

`amazon.opentelemetry.distro.instrumentation.mcp` propagates trace context only in the JSON-RPC params._meta body field by default. HTTP-header propagation is actively suppressed. When the MCP server extracts trace context from HTTP headers (as any non-ADOT-instrumented server does — API Gateway–fronted Lambdas with X-Ray, generic HTTP proxies, non-Python MCP implementations, service meshes, etc.) the request arrives with no traceparent header, and every server-side span becomes an orphan trace disconnected from the client-side trace.

The behavior is controlled by the undocumented env var `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` (default "true", added in #695).

# Environment

  - Package: aws-opentelemetry-distro >= 0.16.0 (native MCP instrumentation), most recently reproduced on 0.18.0
  - MCP SDK: mcp >= 1.10.0
  - Transport: mcp.client.streamable_http.streamablehttp_client
  - Server: AWS API Gateway → Lambda. API Gateway extracts X-Amzn-Trace-Id / W3C traceparent from request headers and forwards to Lambda for X-Ray active tracing. Server does NOT run McpInstrumentor — it's a general-purpose HTTP endpoint.

# What happens

Default configuration (`OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` unset → "true"):

1. Client calls `session.call_tool("foo", {...})`.
2. `ClientWrapper.wrap_session_send` (lines 189–252) creates the mcp tools/call foo span and injects traceparent/tracestate/baggage into params._meta inside the JSON-RPC body.
3. `ClientWrapper.wrap_http_client` (lines 268–293) wraps the streamable-HTTP transport in suppress_http_instrumentation(). All httpx work — including the actual POST to the MCP server — runs under that suppression.
4. The httpx client instrumentation's `TraceContextTextMapPropagator.inject()` never runs, so no traceparent HTTP header is emitted.
5. On the wire, the outbound HTTP request has: content-type, accept, authorization, mcp-session-id, and the JSON body with _meta.traceparent inside — but no traceparent header.
6. The MCP server (API Gateway, in our case) reads only headers. It sees no upstream trace context, generates a fresh trace root, and emits spans on a new trace ID that is not linked to the caller's trace.

Result: every MCP tool call the client makes produces two disconnected traces — a client-side tree that dead-ends at mcp.session/mcp tools/call, and one orphan tree per hop on the server side that starts fresh.

Client-side trace (broken)
```
  InvokeAgentRuntime                                (root)
  └─ POST /invocations
     └─ invoke_agent research-agent
        └─ execute_event_loop_cycle
           └─ chat anthropic.claude-sonnet-4         (Bedrock)
           └─ execute_tool query_metrics
              └─ mcp.session                          ← branch DEAD-ENDS HERE
                 (child spans should live below this — but they land on a different trace ID)
```
Corresponding orphan trace (unlinked)
```
  POST /mcp                                          ← orphan root, no parent
  └─ MetricsService-Frontend/LambdaService
     └─ MetricsService-Frontend/LambdaExecutionEnvironment
```

# Workaround

Set `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION=false` on the client environment. httpx client instrumentation runs normally, traceparent lands in the outgoing HTTP headers, and the server side (API Gateway → Lambda) attaches its spans to the caller's trace.

After trace (workaround applied)
```
InvokeAgentRuntime                                (root, same as before)
└─ POST /invocations
   └─ invoke_agent research-agent
      └─ execute_event_loop_cycle
         └─ chat anthropic.claude-sonnet-4         (Bedrock)
         └─ execute_tool query_metrics
            └─ mcp.session
               └─ mcp tools/call query_metrics
                  └─ POST                          ← NEW (httpx CLIENT)
                     └─ POST /mcp                   (API GW SERVER)
                        └─ MetricsService-Frontend/LambdaService
                           └─ MetricsService-Frontend/LambdaExecutionEnvironment
```
Everything is now on one trace ID.

# Why this is an issue

1. The default breaks propagation for the majority of real-world MCP topologies. The OTel MCP semantic convention (https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/mcp.md) that specifies _meta propagation is currently in development (acknowledged in the [code](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/instrumentation/mcp/_wrappers.py#L86-L87)) and requires the MCP server to be running an implementation that also participates in it.

1. Any MCP server that is not running ADOT MCP instrumentation — API-Gateway-fronted Lambdas, generic HTTP reverse proxies, non-Python MCP servers, servers behind service meshes, custom HTTP frameworks — extracts context exclusively from HTTP headers. For these servers, the default is broken, not "slightly noisy": the client trace and server trace never join.
2. The workaround env var is undocumented. `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` doesn't appear in the README, in CHANGELOG.md outside the PR reference, or anywhere in aws-opentelemetry-distro's docs. Nothing surfaces to a user seeing orphan trace trees that this knob exists.
3. No test covers the non-instrumented-server case. tests/.../test_mcp_instrumentor.py::test_http_span_suppression sets up mcpserver.py which installs McpInstrumentor on the server (line 21) and extracts context via _meta. The test asserts that HTTP spans are absent when suppression is on, which is by design — but the test can never fail on the header-propagation regression because both endpoints understand _meta.
4. The user has no diagnostic breadcrumb. A user seeing orphan traces has to reason from "the wire ought to have traceparent" → "httpx client instrumentation would have added that" → "something must be suppressing it" → grep the distro source → find the env var. That's several hops of source-code archaeology to recover from a silent default. Ideally the distro emits a warning when it detects the topology (server response doesn't include an ADOT-instrumented span header, or _meta propagation is unused).

# Suggested fixes (in preference order)

Always inject headers alongside _meta. The suppression today does both things simultaneously (suppresses the child span AND its header injection). Decouple them: keep suppressing the redundant CLIENT span (that's what PR #695 was actually trying to
solve), but always let the propagator's header injection run. Concretely, either (a) call TraceContextTextMapPropagator().inject(carrier=request.headers) explicitly at the httpx-request-hook layer inside the wrapper, or (b) use set_http_response_hook/request_hook on the httpx instrumentation instead of the blunt suppress_http_instrumentation context.

# Reproduction

Any MCP client using mcp.ClientSession + streamablehttp_client against an MCP server that reads trace context from HTTP headers (not JSON-RPC _meta).

jaredcnance:
## Summary

Even with `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION=true` (the default from #695), inject `traceparent`/`tracestate`/`baggage` into the outgoing HTTP request headers so MCP servers that read trace context only from HTTP headers can join the caller's trace.

Fixes #826.

## Background

The default from #695 correctly deduplicates the redundant HTTP CLIENT span, but it does so via `suppress_http_instrumentation()`, which also short-circuits the HTTP client instrumentation's own header injection — the only mechanism that would otherwise set `traceparent` on the wire. MCP servers that only participate in HTTP-header propagation (API Gateways, service meshes, non-Python MCP servers, any HTTP endpoint without ADOT MCP instrumentation) therefore receive no context and start a fresh trace, silently breaking E2E propagation for a large class of real-world topologies.

#826 has before/after trace evidence from a real AWS deployment showing the client trace dead-ending at `mcp.session` while every server-side hop lands on its own fresh trace ID.

## Design

Compatible with the [OTel MCP semantic convention](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/mcp.md):

> MCP server instrumentation SHOULD, by default, use context extracted from MCP `params._meta` as a parent for MCP server span and SHOULD link current ambient context, if it's present.

The spec designates `_meta` as the MCP parent channel and HTTP context as an orthogonal transport signal (a span link on ADOT-instrumented servers). Nothing in the spec forbids injecting into HTTP headers as well; the existing ADOT server-side wrapper only extracts from `_meta`, so double-parenting on ADOT-instrumented servers is not a risk. This change adds HTTP-header propagation as a compatible fallback for non-ADOT servers while leaving the existing `_meta` propagation and the span-suppression semantics unchanged.

## Implementation

`ClientWrapper.wrap_http_client` now wraps the transport's `httpx_client_factory` (or supplies its own pre-built `http_client` for the newer `streamable_http_client(url, http_client=...)` API) so every `httpx.AsyncClient` the MCP transport uses has a request event hook that calls `self._propagators.inject(carrier=request.headers)`.

The hook takes the OTel context *at request-send time*, so per-tool-call parents (which change during a shared MCP session) are honored correctly. If httpx or MCP is not importable, or the wrapped function signature doesn't match, the patcher no-ops silently.

The change is purely additive with respect to `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION`:

| `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` | HTTP CLIENT span | `traceparent` HTTP header |
| --- | --- | --- |
| `true` (default) — before this PR | suppressed ✓ | **absent ✗** (bug) |
| `true` (default) — after this PR | suppressed ✓ | **present ✓** |
| `false` — before this PR | present | present |
| `false` — after this PR | present | present |

Span suppression, ping suppression, active-parent respect, tool/prompt/resource spans, and error handling across stdio+http+sse transports are unchanged.

## Testing

**New tests** in `test_mcp_instrumentor.py`:

- `test_http_headers_carry_trace_context_regardless_of_suppression` — exercises `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` in `{true, false, default}` and asserts every outgoing MCP HTTP request carries a well-formed W3C `traceparent` header (validates format: `00-<32hex trace-id>-<16hex span-id>-<2hex flags>`).
- `test_http_headers_prebuilt_client_gets_hook` — verifies the hook is installed on a user-supplied pre-built `http_client` (newer `streamable_http_client(url, http_client=...)` API).

**Existing tests** (15 total including the two new ones):
```
$ python -m pytest test_mcp_instrumentor.py
======================== 15 passed, 1 warning in 20.89s ========================
```

**Lint**:
```
$ black --check aws-opentelemetry-distro/src/.../_wrappers.py .../test_mcp_instrumentor.py
All done! ✨ 🍰 ✨
2 files would be left unchanged.

$ flake8 aws-opentelemetry-distro/src/.../_wrappers.py .../test_mcp_instrumentor.py
(no output)

$ pylint --rcfile pyproject.toml aws-opentelemetry-distro/src/.../_wrappers.py
Your code has been rated at 9.01/10 (previous run: 9.01/10, +0.00)
```

## Related PRs

- #683 — first added HTTP context propagation for MCP requests (via `_meta` + `_handle_post_request` restoration)
- #695 — introduced `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` defaulting to `true` (this is the PR whose default silently broke non-ADOT-server topologies)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 ## Unreleased
 
 - fix(mcp-instrumentation): always inject W3C trace context into outbound HTTP request headers so MCP servers that read context only from HTTP (API Gateways, service meshes, non-Python MCP servers) can join the caller's trace, even with `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` enabled
-  ([#826](https://github.com/aws-observability/aws-otel-python-instrumentation/issues/826))
+  ([#827](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/827))
 - fix(crewai): use native per-call token usage when crewai provides it
   ([#822](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/822))
 - fix(crewai): normalize tool description across crewai versions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- fix(mcp-instrumentation): always inject W3C trace context into outbound HTTP request headers so MCP servers that read context only from HTTP (API Gateways, service meshes, non-Python MCP servers) can join the caller's trace, even with `OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION` enabled
+  ([#826](https://github.com/aws-observability/aws-otel-python-instrumentation/issues/826))
 - fix(crewai): use native per-call token usage when crewai provides it
   ([#822](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/822))
 - fix(crewai): normalize tool description across crewai versions
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/instrumentation/mcp/__init__.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/instrumentation/mcp/__init__.py
@@ -76,6 +76,11 @@ class McpInstrumentor(BaseInstrumentor):
             "StreamableHTTPTransport._handle_post_request",
             self._client_wrapper.wrap_handle_post_request,
         )
+        try_wrap(
+            _HTTP_MODULE,
+            "StreamableHTTPTransport._prepare_headers",
+            self._client_wrapper.wrap_prepare_headers,
+        )
 
         _LOG.debug("Instrument MCP server ASGI apps to suppress redundant HTTP spans.")
 
@@ -114,6 +119,10 @@ class McpInstrumentor(BaseInstrumentor):
                 try_unwrap(
                     streamable_http.StreamableHTTPTransport,
                     "_handle_post_request",
+                )
+                try_unwrap(
+                    streamable_http.StreamableHTTPTransport,
+                    "_prepare_headers",
                 )
             try_unwrap(fastmcp_server.FastMCP, "streamable_http_app")
             try_unwrap(fastmcp_server.FastMCP, "sse_app")
--- a/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/instrumentation/mcp/_wrappers.py
+++ b/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/instrumentation/mcp/_wrappers.py
@@ -1,7 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import inspect
 import logging
 import os
 from contextlib import asynccontextmanager
@@ -43,43 +42,6 @@ OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION = "OTEL_MCP_SUPPRESS_HTTP_INSTRUMENTATION
 
 # Context key for storing client transport metadata alongside the session span.
 _TRANSPORT_KEY = context.create_key("mcp_client_transport")
-
-
-def _get_parameter_names(func: Callable[..., Any]) -> Tuple[str, ...]:
-    """Return the parameter names of ``func`` (best-effort, empty on failure)."""
-    try:
-        return tuple(inspect.signature(func).parameters.keys())
-    except (TypeError, ValueError):
-        return ()
-
-
-# Sentinel attribute set on httpx.AsyncClient instances to prevent attaching the
-# propagation hook more than once (closures compare unequal by identity, so a
-# plain ``not in`` list check is insufficient for dedup).
-_HOOK_INSTALLED_ATTR = "_adot_mcp_propagation_hook_installed"
-
-
-def _attach_propagation_hook(client: Any, propagators: Any) -> None:
-    """Attach a request event hook that injects W3C trace context into HTTP headers.
-
-    Idempotent: marks the client with a sentinel attribute to prevent
-    accumulating duplicate hooks across sessions that reuse the same client.
-    """
-    try:
-        if getattr(client, _HOOK_INSTALLED_ATTR, False):
-            return
-
-        async def _inject_trace_context(request: Any) -> None:
-            try:
-                propagators.inject(carrier=request.headers)
-            except Exception:  # pylint: disable=broad-exception-caught
-                _LOG.debug("MCP trace-context header injection failed", exc_info=True)
-
-        request_hooks = client.event_hooks.setdefault("request", [])
-        request_hooks.append(_inject_trace_context)
-        setattr(client, _HOOK_INSTALLED_ATTR, True)
-    except Exception:  # pylint: disable=broad-exception-caught
-        _LOG.debug("Failed to attach MCP trace-context injection hook", exc_info=True)
 
 
 class McpWrapper:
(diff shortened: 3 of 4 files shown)


jaredcnance:
Let’s hold on merging this until next week as I want to run a few more e2e tests. Will post test results to the PR.

jaredcnance:
Verified end-to-end on an AWS deployment where the MCP server is a Lambda Java service fronted by API Gateway with X-Ray active tracing. Before this change: client trace dead-ended at `mcp.session`, and each MCP tool call produced an orphan trace on the server side (fresh root at `POST /mcp`). After this change: 

```
  [SERVER] AgentCore.Runtime.Invoke                       ← agent runtime root
  └─ [INTERNAL] execute_event_loop_cycle                  (agent loop)
     ├─ [INTERNAL] chat
     │  └─ [CLIENT] chat anthropic.claude-sonnet-4        (Bedrock)
     └─ [INTERNAL] execute_tool query_metrics             (MCP tool)
        └─ [CLIENT] mcp tools/call query_metrics             (httpx → traceparent header)
           └─ [SERVER] MetricsApi/prod                    (AWS::ApiGateway::Stage)
              └─ [CLIENT] Lambda                          (API GW integration)
                 └─ [SERVER] MetricsService-Frontend/LambdaService
                    └─ [SERVER] MetricsService-Frontend/LambdaExecutionEnvironment
```


Sample client span:

```
  {
    "name": "mcp tools/call query_metrics",
    "kind": "CLIENT",
    "traceId": "5f369767e659917c65a1a4573210866a",
    "spanId": "bb63e5278111c0c5",
    "parentSpanId": "74d96fd91d70c4e8",
    "durationMs": 15.3,
    "status": { "code": "OK" },
    "attributes": {
      "mcp.method.name": "tools/call",
      "mcp.protocol.version": "2025-11-25",
      "mcp.session.id": "e9a00860bd084ee88d80e6e990f6e3fd",
      "jsonrpc.request.id": "1",
      "gen_ai.tool.name": "query_metrics",
      "gen_ai.operation.name": "execute_tool",
      "gen_ai.tool.call.arguments": "{\"namespace\": \"AWS/Lambda\", \"metric_name\": \"Errors\"}",
      "gen_ai.tool.call.result": "…{\"datapoints\": [{\"timestamp\": \"2026-07-19T00:00:00Z\", \"value\": 42.0}]}… isError=False",
      "network.transport": "tcp",
      "server.address": "127.0.0.1",
      "server.port": 54499
    },
    "instrumentationScope": {
      "name": "amazon.opentelemetry.distro.instrumentation.mcp",
      "version": "0.18.0.dev0"
    }
  }
```

liustve:
Thanks @jaredcnance for the changes and testing! 
One suggestion: I think we should follow [OTel botocore instrumentation pattern](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L226) and wrap the header-prep function.

The MCP function to wrap would be is [StreamableHTTPTransport._prepare_headers](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/streamable_http.py#L113), which already runs under our attached _meta context.

under `wrapper.py`
```
def wrap_prepare_headers(self, wrapped, instance, args, kwargs):
    headers = wrapped(*args, **kwargs)
    if self._should_suppress_http_spans:  # else httpx injects it itself
        self._propagators.inject(carrier=headers)
    return headers
```

under _instrument() and _uninstrument() in:

```
try_wrap(
    _HTTP_MODULE,
    "StreamableHTTPTransport._prepare_headers",
    self._client_wrapper.wrap_prepare_headers,
)
```

```
if hasattr(streamable_http, "StreamableHTTPTransport"):
    try_unwrap(streamable_http.StreamableHTTPTransport, "_prepare_headers")
```

jaredcnance:
sweet -- thanks for the pointers. PR is updated, let me know if this makes sense. verified end-to-end in deployed environment.

liustve (approved):

mxiamxia (approved):

PR #851, opened 2026-08-07
feat: span metrics for cloudwatch-plugin-otel (connector, instrumentor, always-record sampler)

liustve:
Adds span metrics to `cloudwatch-plugin-otel`: a `SpanMetricsConnector` that derives call-count and duration metrics from every span, an `AlwaysRecordSampler` so spans the root sampler would drop are still counted without being exported, and a `SpanMetricsInstrumentor` that wires both in automatically under `opentelemetry-instrument`. Covered by unit tests across the Python 3.10–3.14 × oldest/latest OTel matrix, including an end-to-end auto-instrumentation test.


cloudwatch-plugin-otel/src/plugins/opentelemetry/cloudwatch/span_metrics/_constants.py
```diff
@@ -0,0 +1,37 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Final
+
+
+class _SpanMetrics:
+    SCOPE_NAME: Final = "cloudwatch.plugin.otel.span-metrics"
```
srprash:
```suggestion
    SCOPE_NAME: Final = "cloudwatch.plugin.otel.span_metrics"
```

srprash: as per OTel snake case convention.

cloudwatch-plugin-otel/src/plugins/opentelemetry/cloudwatch/span_metrics/connector.py
```diff
@@ -0,0 +1,169 @@
...
+            span_attributes,
+            DB_COLLECTION_NAME,
+            DB_SQL_TABLE,
+            DB_MONGODB_COLLECTION,
+            DB_CASSANDRA_TABLE,
+            DB_COSMOSDB_CONTAINER,
+        )
+        self._copy(attributes, span_attributes, MESSAGING_SYSTEM)
+        self._copy(attributes, span_attributes, MESSAGING_OPERATION_NAME)
+        self._copy(attributes, span_attributes, MESSAGING_OPERATION_TYPE, MESSAGING_OPERATION)
```
srprash: we don't need these attributes on the span metrics as per the internal spec.

srprash:
Please update the readme as well

liustve:
Addressed the review feedback in de92ad3: the instrumentation scope now uses `cloudwatch.plugin.otel.span_metrics`, unsupported messaging operation attributes are no longer copied, and the README documents span metrics under its own section while keeping the package overview feature-neutral.

cloudwatch-plugin-otel/src/plugins/opentelemetry/cloudwatch/span_metrics/connector.py
liustve: fixed

cloudwatch-plugin-otel/src/plugins/opentelemetry/cloudwatch/span_metrics/_constants.py
liustve: fixed

cloudwatch-plugin-otel/src/plugins/opentelemetry/cloudwatch/span_metrics/connector.py
```diff
@@ -0,0 +1,166 @@
...
+        except Exception:
+            _logger.debug("Failed to stamp span metrics attributes", exc_info=True)
+
+    @override
+    def on_end(self, span: ReadableSpan) -> None:
+        if not self.enabled:
+            return
+        try:
+            attributes = self._build_metric_attributes(span)
+            duration_seconds = (span.end_time - span.start_time) / _SpanMetrics.NANOS_PER_SECOND
```
srprash: In `on_end`, the duration is computed *before* `_calls_counter.add(1, ...)`, and both are inside the same `try/except`:

```python
duration_seconds = (span.end_time - span.start_time) / _SpanMetrics.NANOS_PER_SECOND
self._calls_counter.add(1, attributes)
self._duration_histogram.record(duration_seconds, attributes)
```

If `span.end_time` (or `start_time`) is `None`, line 1 raises `TypeError`, which the broad `except Exception` swallows (logged only at `debug`). Because the subtraction runs before `add(1)`, **both** the `calls` count and the `duration` sample are dropped for that span — the call is silently uncounted, which undercuts the "100% of spans" guarantee with no visible signal.

Suggestion: increment the counter first, then derive/record duration, so a bad/missing timestamp can at most drop the duration sample, never the call count:

```python
self._calls_counter.add(1, attributes)
if span.end_time is not None and span.start_time is not None:
    duration_seconds = (span.end_time - span.start_time) / _SpanMetrics.NANOS_PER_SECOND
    self._duration_histogram.record(duration_seconds, attributes)
```

Worth a test for the `end_time is None` path asserting `calls` still increments (the current `test_malformed_span_never_raises` fails earlier at attribute build and asserts zero metrics, so it doesn't cover this).

liustve: fixed

srprash (approved):

PR #852, opened 2026-08-07
Python eks parallel pipeline

haneric00:
EKS tests for ADOT Python currently run sequentially, dedicating almost an hour to EKS testing on every build. This PR wires ADOT Python's main build up to the new parallel EKS test workflows (added in the companion aws-application-signals-test-framework PR), running all supported Python versions concurrently against one shared cluster and removing a major bottleneck in the build/release pipeline.

The existing sequential EKS test is untouched; this only changes how application-signals-e2e-test.yml invokes EKS testing.

Description of changes:

Rework the EKS portion of .github/workflows/application-signals-e2e-test.yml from a single sequential test into a parallel flow built on the framework's new reusable workflows. The flow is:

service-events-eks  ->  eks-setup  ->  eks-py3.10 … eks-py3.14 (parallel)  ->  eks-cleanup

- service-events-eks runs first. It uses the unchanged shared Service Events workflow, which installs and tears down App Signals itself and binds a fixed NodePort, so it must not overlap the parallel version tests. Placing it upstream (rather than after cleanup) means re-running eks-setup re-runs the whole version flow without dragging Service Events along.
- eks-setup installs App Signals on the shared cluster once, before the version tests. It runs regardless of whether Service Events passed or failed (so a Service Events failure can't block the version matrix) while still respecting run cancellation.
- eks-py310…eks-py314 each call the new python-eks-shared-cluster-test.yml for one Python version. They run in parallel against the shared cluster once setup succeeds.
- eks-cleanup tears down the shared App Signals state once, after every version test finishes. Teardown still runs even if a version test fails.

Update the validate-e2e-tests-are-accounted-for exclusions to python-eks-test.yml,eks-setup-app-signals.yml,eks-cleanup-app-signals.yml — this repo uses the new shared-cluster test instead of the sequential python-eks-test.yml, and setup/cleanup are lifecycle workflows, not tests.

Re-running a failed EKS run: because the version tests depend on eks-setup's shared cluster state (which eks-cleanup tears down), re-running a single failed version job on its own won't work. Instead, use "Re-run this job" on eks-setup, which cascades to re-run setup → version tests → cleanup. The shared-cluster test prints this guidance on failure.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/.github/workflows/application-signals-e2e-test.yml
+++ b/.github/workflows/application-signals-e2e-test.yml
@@ -366,16 +366,6 @@ jobs:
   # - Update the `validate-e2e-tests-are-accounted-for` job to change which "workflow files are expected to be used by this repo"
   #   (see: https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/validate-e2e-tests-are-accounted-for.yml)
   validate-all-tests-are-accounted-for:
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/validate-e2e-tests-are-accounted-for.yml@main
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/validate-e2e-tests-are-accounted-for.yml@python-eks-parallel-pipeline
     with:
-      # The validator strict-equality-compares TESTS_EXPECTED (framework files matching python-*test.yml)
-      # against TESTS_ACTUALLY_USED (every framework workflow this file `uses:`). Exclusions are applied
-      # to BOTH sides, so anything not in both lists must be excluded:
-      #   - python-eks-test.yml       : sequential EKS workflow. Expected (matches python-*test.yml) but
-      #                                 this repo uses python-eks-parallel-test.yml instead, so it is
-      #                                 never in USED -> exclude to drop it from EXPECTED.
-      #   - eks-setup-app-signals.yml : used here, but does NOT match python-*test.yml, so never in
-      #   - eks-cleanup-app-signals.yml EXPECTED -> exclude to drop them from USED.
-      # (python-eks-parallel-test.yml itself needs NO exclusion: it matches python-*test.yml AND is used,
-      #  so it is in both lists once the framework branch is merged.)
       exclusions: 'python-eks-test.yml,eks-setup-app-signals.yml,eks-cleanup-app-signals.yml'
--- a/.github/workflows/application-signals-e2e-test.yml
+++ b/.github/workflows/application-signals-e2e-test.yml
@@ -125,19 +125,98 @@ jobs:
       staging-wheel-name: ${{ inputs.staging-wheel-name }}
 
 
-  # --- Python EKS E2E (setup -> parallel versions -> cleanup) -----------------------------------
-  # The entire EKS flow lives in one reusable workflow (python-eks-e2e.yml) so it re-runs as a
-  # single unit. Re-running just this job (e.g. "Re-run failed jobs") re-executes setup -> eks-py* ->
-  # cleanup together, INCLUDING the shared-cluster setup the version tests depend on, WITHOUT
-  # re-running the rest of this suite. service-events-eks still runs after, on the cold cluster.
-  eks-e2e:
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/python-eks-e2e.yml@python-eks-parallel-pipeline
+  # --- Python EKS pipeline: service-events (first) -> setup -> parallel versions -> cleanup ------
+  # service-events-eks runs FIRST on a cold cluster: the UNCHANGED shared SE workflow self-installs
+  # and tears down App Signals and binds NodePort 30100, so it must not overlap the parallel eks-py*
+  # jobs. eks-setup waits for SE (needs) but runs regardless of SE's result (if: !cancelled()), so a
+  # SE failure can't block the version tests. Keeping SE UPSTREAM means re-running the eks-setup job
+  # re-runs setup -> eks-py* -> cleanup WITHOUT dragging service-events along.
+  service-events-eks:
+    if: ${{ always() }}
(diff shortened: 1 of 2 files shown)


.github/workflows/application-signals-e2e-test.yml
```diff
@@ -124,14 +124,21 @@ jobs:
...
+    secrets: inherit
+    with:
+      test-cluster-name: 'e2e-python-adot-test'
+      aws-region: us-east-1
 
   eks-py310-amd64:
-    if: ${{ always() }}
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/python-eks-test.yml@main
+    needs: eks-setup
+    if: ${{ always() && needs.eks-setup.result == 'success' }}
```
jj22ee: Is `always()` redundant?

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -124,14 +124,21 @@ jobs:
...
+
+  # --- Parallel Python EKS pipeline -------------------------------------------------------------
+  # Shape: eks-setup -> [eks-py310..314 concurrently] -> eks-cleanup -> service-events-eks
+
+  eks-setup:
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/eks-setup-app-signals.yml@main
+    secrets: inherit
+    with:
+      test-cluster-name: 'e2e-python-adot-test'
+      aws-region: us-east-1
```
jj22ee, line 147: One consequence of having the setup done outside an e2e test, like `eks-py310-amd64`, is that we will never be able to "re-run" failed jobs like `eks-py310-amd64` individually, as they will fail because they'll skip the `eks-setup` job while the `eks-cleanup` job already ran. To re-run the tests, we have to run this entire workflow which means re-testing all python versions. This might be confusing too as I believe we need to `click re-run` on the `eks-setup` job instead of the `eks-py310-amd64` job to re-run the `eks-py310-amd64` job.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -165,9 +172,9 @@ jobs:
       python-version: '3.12'
 
   eks-py313-amd64:
-    if: ${{ always() }}
-    needs: eks-py312-amd64
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/python-eks-test.yml@main
+    needs: eks-setup
+    if: ${{ always() && needs.eks-setup.result == 'success' }}
```
jj22ee: The `if:` statement seems like it can be removed and we get the same functionality? Same comment for the other python eks tests.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -337,3 +356,5 @@ jobs:
   #   (see: https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/validate-e2e-tests-are-accounted-for.yml)
   validate-all-tests-are-accounted-for:
     uses: aws-observability/aws-application-signals-test-framework/.github/workflows/validate-e2e-tests-are-accounted-for.yml@main
+    with:
+      exclusions: 'python-eks-test.yml,eks-setup-app-signals.yml,eks-cleanup-app-signals.yml'
```
jj22ee, line 350: You'll also need to update the `exclusions` list in https://github.com/aws/amazon-cloudwatch-agent/blob/main/.github/workflows/application-signals-e2e-test.yml#L269-L270

.github/workflows/application-signals-e2e-test.yml
jj22ee: I recall you mentioned the reason we extracted `eks-cleanup` out of each individual e2e test is so we can do a `global eks-cleanup` for all of them at once, because doing them individually would allow one cleanup to cause issues in another ongoing test.

Is it still possible for each individual test to have their own `eks-setup` still instead of having this `global eks-setup`? And was it okay when concurrent `eks-setup` runs, or did that cause issues previously too? I'm thinking that if `eks-setup` stays in each e2e test workflow, we'd still be able to re-run an individual test.

haneric00: Hi Jonathan, thanks for taking an in depth look!

I do believe both setup and cleanup need to be separated. For setup there is a race condition in the enable-app-signals.sh script that prevented IRSA permissions from being added to the CW agent once it is online.

jj22ee: I think it's likely one of us will not know this tradeoff in the future, and then attempt to re-run a failed workflow which is guaranteed to fail. This would be another problem we'd need to work on.

In order to improve the user-experience, I'm thinking of a couple ideas:
1.  Make it more intuitive that if someone wants to re-run the tests, they do not run the version-specific individual e2e test?
    - Right now, the `setup` workflow is needed to actually run the tests, but it's faraway from the `python-eks` tests, so naturally we'd still try to re-run the `python-eks` test instead of the `setup` workflow. Also due to the naming, we'd re-run `python-eks` instead of `setup`. We also have other eks tests that do not rely on `setup` either which can be run individually.
    - Instead of having `setup/cleanup/python-eks` tests be child workflows of `application-signals-e2e-test.yml`, you can have the `setup/cleanup/python-eks` tests be child workflows of a `setup and run parallel eks tests` workflow that is a child of `application-signals-e2e-test.yml`. I think this way, all the `setup/cleanup/python-eks` jobs will be bundled together in the UI. Then you can rename these to be more intuitive so we know which workflow to use if we want to re-run it, for example change `setup -> setup-and-initiate-tests`, and `python-eks -> trigger-telemetry-generation-and-validate` 
2. What if we don't have multiple workflows/jobs, just do what we did previously, but have all parallelization of the `sample-app deployments` and `python-version validations` be done by using Built-in Bash Backgrounding (`&` and `wait`)?

image

v-iw (dismissed):

srprash (dismissed):

srprash (approved):

PR #855, opened 2026-08-14
Extract shared Python release actions

liustve:
Extracts the existing main-build gate, PyPI publishing steps, and common Python release-note preparation into reusable composite actions. The bodies are moved from the current release workflow with only path/input/output plumbing; package-specific downloads, release text, checksums, and assets remain in the callers. Equivalence commands and results are in the verification comment.

### DIFF
--- a/.github/actions/get_main_build_run/action.yml
+++ b/.github/actions/get_main_build_run/action.yml
@@ -27,30 +27,20 @@ runs:
       id: main_build
       shell: bash
       env:
-        GH_TOKEN: ${{ inputs.github-token }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
         TARGET_BRANCH: ${{ inputs.branch }}
         WORKFLOW_NAME: ${{ inputs.workflow-name }}
       run: |
-        set -euo pipefail
-
         WORKFLOW_ID=$(gh api "repos/$GITHUB_REPOSITORY/actions/workflows" |
-          jq -r --arg name "$WORKFLOW_NAME" 'first(.workflows[] | select(.name == $name) | .id) // empty')
-        if [ -z "$WORKFLOW_ID" ]; then
-          echo "Workflow '$WORKFLOW_NAME' was not found."
-          exit 1
-        fi
+          jq -r --arg name "$WORKFLOW_NAME" '.workflows[] | select(.name == $name) | .id')
+        LATEST_RUN=$(gh api "repos/$GITHUB_REPOSITORY/actions/workflows/$WORKFLOW_ID/runs" |
+          jq -c --arg branch "$TARGET_BRANCH" \
+            '[.workflow_runs[] | select(.head_branch == $branch)] | sort_by(.created_at) | .[-1]')
+        STATUS=$(echo "$LATEST_RUN" | jq -r '.status')
+        CONCLUSION=$(echo "$LATEST_RUN" | jq -r '.conclusion')
+        RUN_ID=$(echo "$LATEST_RUN" | jq -r '.id')
 
-        LATEST_RUN=$(gh api --method GET "repos/$GITHUB_REPOSITORY/actions/workflows/$WORKFLOW_ID/runs" \
-          -f branch="$TARGET_BRANCH" \
-          -f per_page=1)
-        RUN_ID=$(echo "$LATEST_RUN" | jq -r '.workflow_runs[0].id // empty')
-        STATUS=$(echo "$LATEST_RUN" | jq -r '.workflow_runs[0].status // empty')
-        CONCLUSION=$(echo "$LATEST_RUN" | jq -r '.workflow_runs[0].conclusion // empty')
-
-        if [ -z "$RUN_ID" ]; then
-          echo "No main build was found on branch $TARGET_BRANCH."
-          exit 1
-        elif [ "$STATUS" != "completed" ]; then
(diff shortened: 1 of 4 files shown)


liustve:
Verification performed against `origin/main`.

### Why these extraction boundaries

The actions follow the existing release workflow's reusable units:

- `get_main_build_run` owns only the shared green-build lookup and returns the run ID. Artifact downloads stay in each workflow because the distro downloads a wheel plus Lambda layer while the plugin downloads its own wheel plus sdist.
- `publish_pypi` owns the exact four shared steps: download wheel, download sdist, publish to TestPyPI, then publish to PyPI.
- `prepare_python_release_notes` owns the exact shared dependency/CHANGELOG extraction and the common `What's Changed` / `Upstream Components` prefix.

Package-specific release text, checksums, tags, and attached assets remain in the callers. Abstracting those would require distro/plugin switches and would hide meaningful differences rather than remove duplication.

### Main-build gate

These run the original selector and the parameterized selector from the composite action against the same live API response:

```bash
gh api repos/aws-observability/aws-otel-python-instrumentation/actions/workflows/main-build.yml/runs \
  --jq '[.workflow_runs[] | select(.head_branch=="main")] | sort_by(.created_at) | .[-1] | [.id, .status, .conclusion] | @tsv'

gh api repos/aws-observability/aws-otel-python-instrumentation/actions/workflows/main-build.yml/runs |
  jq -r --arg branch main \
    '[.workflow_runs[] | select(.head_branch == $branch)] | sort_by(.created_at) | .[-1] | [.id, .status, .conclusion] | @tsv'
```

Both returned:

```text
31743522232	completed	failure
```

The action keeps the original endpoints, branch filter, `created_at` sort, queued/in-progress rejection, and `conclusion == success` requirement.

### PyPI publication

This expands the composite-action defaults and caller inputs, then compares every extracted `uses:` and `with:` value with the four original workflow steps:

```bash
ruby -ryaml -e '
old_workflow = YAML.load(`git show origin/main:.github/workflows/release-build.yml`)
new_workflow = YAML.load_file(".github/workflows/release-build.yml")
action = YAML.load_file(".github/actions/publish_pypi/action.yml")
relevant = ->(step) { step["uses"]&.match?(%r{^(actions/download-artifact|pypa/gh-action-pypi-publish)@}) }
old_steps = old_workflow.fetch("jobs").fetch("publish-sdk").fetch("steps").select(&relevant)
call = new_workflow.fetch("jobs").fetch("publish-sdk").fetch("steps").find { |step| step["uses"] == "./.github/actions/publish_pypi" }
values = action.fetch("inputs").transform_values { |input| input["default"] }.merge(call.fetch("with"))
new_steps = action.fetch("runs").fetch("steps").select(&relevant).map do |step|
  normalized_with = step.fetch("with", {}).transform_values do |value|
    value.is_a?(String) ? value.gsub(/\$\{\{ inputs\.([^ }]+) \}\}/) { values.fetch(Regexp.last_match(1)) } : value
  end
  { "uses" => step["uses"], "with" => normalized_with }
end
old_steps = old_steps.map { |step| { "uses" => step["uses"], "with" => step.fetch("with", {}) } }
abort "PyPI extraction changed behavior" unless old_steps == new_steps
puts "Equivalent: #{new_steps.length} action invocations; all uses/with values match"
'
```

Output:

```text
Equivalent: 4 action invocations; all uses/with values match
```

### Release-note preparation

I ran the original inline dependency/CHANGELOG block and the extracted action with the same distro inputs:

```bash
PYPROJECT_FILE=aws-opentelemetry-distro/pyproject.toml
CHANGELOG_FILE=CHANGELOG.md
VERSION=0.19.0
diff -u /private/tmp/release_notes_old.md /private/tmp/release_notes_new.md
```

`diff -u` produced no output: the generated common release-note prefix was byte-for-byte identical. The action only parameterizes the pyproject path, changelog path, and version.

### Focused workflow diff

```bash
git diff --color=never --unified=3 origin/main...cloudwatch-release-abstractions -- .github/workflows/release-build.yml
```

The focused diff shows each inline block replaced at the same location by its local action call:

```diff
-      - name: Get main build run ID and download artifacts
+      - name: Get successful main build run
+        uses: ./.github/actions/get_main_build_run

-      - name: Download SDK wheel artifact
-      - name: Download SDK source artifact
-      - name: Publish to TestPyPI
-      - name: Publish to PyPI
+      - name: Publish Python package
+        uses: ./.github/actions/publish_pypi

-          DEPS=$(python3 -c "...")
-          CHANGELOG_ENTRIES=$(python3 -c "...")
-          cat > release_notes.md << EOF
+      - name: Prepare Python release notes
+        uses: ./.github/actions/prepare_python_release_notes
```

The moved bodies are in:

```text
.github/actions/get_main_build_run/action.yml
.github/actions/publish_pypi/action.yml
.github/actions/prepare_python_release_notes/action.yml
```

liustve:
Thanks for the re-review. To clarify the scope: these actions exist only because the same release blocks are now used by both the distro and plugin workflows. This is an extraction-only change—the env inputs and run-ID output are the minimum plumbing needed to move the copied blocks. Pagination, no-run handling, status checks, and publish behavior are intentionally unchanged. The release-note split is only at the common/package-specific boundary and was byte-compared with the original output. I also restored the two original TestPyPI/PyPI comments noted above. No additional hardening or behavioral refactor is intended in this PR.

srprash (approved):

PR #861, opened 2026-08-20
Drop service.name from span-metric datapoint attributes; emit calls with {call} unit

srprash:
## Description

Two changes to the span-metrics schema emitted by the CloudWatch plugin:

1. **`service.name` is no longer a datapoint attribute.** The metrics are recorded into the host SDK's `MeterProvider`, whose resource already carries `service.name`, so duplicating it on every datapoint adds a redundant dimension with no extra information. Consumers read it from the metric resource. This intentionally diverges from the collector spanmetrics connector, which flattens `service.name` into datapoint attributes because collector-side consumers may drop the resource; here the resource is preserved end to end.

2. **`traces.span.metrics.calls` now uses the `{call}` unit** (UCUM annotation) instead of an unset unit, matching OTel semantic-convention counter conventions (cf. `{request}`, `{operation}`).

## Testing

- Unit tests updated: the connector suite asserts `service.name` is never emitted as a datapoint attribute (including when the span's resource carries it), that the metric **resource** carries `service.name`, and that the calls counter uses the literal `{call}` unit. 83 tests pass against both the oldest and latest supported OpenTelemetry requirement sets.
- Contract tests updated: the span-metrics harness now asserts `service.name` is absent from datapoints, present on the metric resource **with the configured value** (key presence alone can never fail — the SDK defaults `service.name` unconditionally), and the `{call}` unit. All 6 span-metrics contract tests (auto / manual / global instrumentation) pass locally against the built wheel and images.

CHANGELOG and README updated accordingly.

### DIFF
--- a/cloudwatch-plugin-otel/CHANGELOG.md
+++ b/cloudwatch-plugin-otel/CHANGELOG.md
@@ -10,6 +10,12 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- `service.name` is no longer emitted as a metric datapoint attribute; it is carried by the
+  metric's resource (the host SDK's resource). Consumers reading `service.name` from datapoint
+  dimensions must read it from the resource instead.
+- The `traces.span.metrics.calls` counter now uses the `{call}` unit (UCUM annotation) instead of
+  an unset unit.
+
 ## v0.1.0 - 2026-08-14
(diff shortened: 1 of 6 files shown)


liustve (approved):

srprash:
Blocking test failures are unrelated to this PR. Bypassing the rule and merging.

PR #869, opened 2026-08-30
Adding artifact verification post-build in the release workflow

haneric00:
Java has post build artifact verification, something very easy to do as Java deals with .jar files. In python this task is a little bit more complicated.

We add:
- a new shell script that is inspired by the original on ADOT java that performs artifact verification
- calls to the new script in the release build workflow
- a new workflow for testing that the shell script works as intended
- trivy artifact scanning to release-build (unrelated but should be added)


.github/workflows/release-build.yml
```diff
@@ -160,6 +203,30 @@ jobs:
...
+        shell: bash
+        env:
+          DOCKER_DEFAULT_PLATFORM: linux/arm64
+        run: .github/scripts/test-adot-python-image.sh "${{ env.RELEASE_PUBLIC_REPOSITORY }}:v${{ env.VERSION }}" "${{ env.VERSION }}"
+
+      - name: Perform image scan
+        uses: ./.github/actions/image_scan
+        with:
+          image-ref: ${{ env.RELEASE_PUBLIC_REPOSITORY }}:v${{ env.VERSION }}
+          severity: 'CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN'
```
v-iw: P1: Can the Trivy scan be its own PR? You flagged it as unrelated, and I think it needs its own discussion. As written it's CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN with exit-code: 1 (hardcoded in image_scan), against daily-scan.trivyignore.yaml whose only two suppressions expired 2026-06-01 (
daily-scan.trivyignore.yaml
, lines 13–19) — so effectively no live suppressions. That means any new LOW/UNKNOWN in python:3.11 blocks a release. daily-scan.yml splits the two severity bands into separate non-coupled steps; should release do the same, or report status instead of blocking?

.github/scripts/test-adot-python-image.sh
```diff
@@ -0,0 +1,133 @@
...
+#   TEST_TAG         image ref to test (a locally built, not-yet-pushed image)
+#   EXPECTED_VERSION optional; when set (release runs pass env.VERSION) the ported distro's
+#                    __version__ must match exactly. Omit for local runs against source.
+
+set -x -e -u
+
+TEST_TAG=$1
+EXPECTED_VERSION="${2:-}"
+
+VOLUME=operator-volume
```
v-iw: P2: VOLUME=operator-volume is a fixed name. docker volume create on an existing volume succeeds and keeps its contents, so a run killed before the trap cleanup EXIT fires (cancelled workflow, OOM) leaves state behind and the next run's diff -r compares a mixed tree. Suggest a per-run unique name, like line 28.

.github/workflows/release-build.yml
```diff
@@ -149,6 +150,48 @@ jobs:
...
           packages-dir: dist-pypi
 
+      # QEMU lets this amd64 runner build and run the arm64 image so both released
+      # architectures are verified before publishing.
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 #3.6.0
+
+      # Build each released architecture locally (not pushed) and run the post-build
+      # integrity + ported-distro verification against each, so a bad artifact never
+      # reaches ECR. amd64 runs natively; arm64 runs under QEMU emulation. DOCKER_DEFAULT_PLATFORM
```
v-iw: P3, trivial: The "a bad artifact never reaches ECR" comment isn't quite accurate. The published-image verify and the scan both run after Build and push image, which pushes to public + private. On failure the image is already live at :v<VERSION> with no untag/rollback. Downstream jobs are correctly blocked, but the comment overstates it. Worth either a cleanup-on-failure step or a reworded comment

v-iw (commented):
Added some comments.

bjrara (approved):

PR #877, opened 2026-09-03
fix: stop logging noisy "Invalid key/value pair (xrsr, None)" warning on span creation

### LINKED ISSUE #874: Noisy logger warning
Hello, since aws-opentelemetry-distro>=0.18.0, we get these logs which are quite noisy. Every trace creation emits these logs:

```
{"timestamp": "2026-09-02T13:43:01Z", "level": "WARNING", "message": "Invalid key/value pair (xrsr, None) found.", "logger": "opentelemetry.trace.span", ...}
```

The key `xrsr` can be traced back to this class:
https://github.com/aws-observability/aws-otel-python-instrumentation/blob/1066621515b032e91222213a3be5859573103b42/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/sampler/_aws_sampling_result.py#L41-L43

This seems to happen only in aiokafka applications. We have other applications with the same aws distro version where these logs do not appear.

Python 3.13.14
AWS ECS Fargate Task (1.14.0 platform version)
Cloudwatch Agent Sidecar (latest version 1.300072.0b1766)
Faststream Application with AIOKafka, instrumented by Faststream itself (latest)

feel free to ask for more information.

srprash:
## Description

Fixes #874.

Since `aws-opentelemetry-distro>=0.18.0`, a `WARNING` is logged on **every span creation** in some apps:

```
WARNING:opentelemetry.trace.span:Invalid key/value pair (xrsr, None) found.
```

### Root cause

`_AwsSamplingResult` unconditionally writes the X-Ray sampling rule hash into the trace state:

```python
if self.trace_state.get(self.AWS_XRAY_SAMPLING_RULE_TRACE_STATE_KEY) is None:
    self.trace_state = self.trace_state.add(self.AWS_XRAY_SAMPLING_RULE_TRACE_STATE_KEY, sampling_rule_hash)
```

When `sampling_rule_hash` is `None`, `TraceState.add("xrsr", None)` fails validation in the OTel SDK, logs the warning, and returns the state **unchanged** — so the write is already a no-op; only the log line is a side effect.

`sampling_rule_hash` is `None` in `_RuleCache.should_sample` when a span is created under a **valid local parent whose trace state carries no `xrsr` key** and no baggage entry (`_rule_cache.py`): `rule_name_to_propagate` becomes `None`, and `self._rule_to_hash_map.get(None, upstream_matched_rule_hash)` resolves to `None`. This matches the reporter's observation that it only reproduces in aiokafka/FastStream apps, where child spans have a valid parent but the `xrsr` hash isn't propagated over the messaging context.

### Fix

Only call `add` when we actually have a hash. Behavior-preserving — the `None` add was already a no-op:

```python
if sampling_rule_hash is not None and self.trace_state.get(self.AWS_XRAY_SAMPLING_RULE_TRACE_STATE_KEY) is None:
    self.trace_state = self.trace_state.add(self.AWS_XRAY_SAMPLING_RULE_TRACE_STATE_KEY, sampling_rule_hash)
```

## Testing

New `test_aws_sampling_result.py` with 4 cases: hash added when present, existing key not overwritten, `None` hash adds no key, and `None` hash emits no warning. All pass; existing `test_rule_cache` suite (14) still green.


wangzlei (approved):

srprash:
Blocking test failures are unrelated to this change. Bypassing the rules and merging.

PR #879, opened 2026-09-03
change fable to opus in langchain tests

haneric00:
change fable to opus in the langchain unit tests because the new release that adds fable 5.1 support also added if checks that specifically target fable 5, you are no longer allowed to use top_k, temperature, top_p attributes if using fable 5 (which we do in our tests) so the easiest fix is to simply move our mock calls to use another model.


liustve (approved):

liustve:
relevant changelog: https://github.com/langchain-ai/langchain/releases/tag/langchain-anthropic%3D%3D1.7.1

syed-ahsan-ishtiaque (approved):

PR #880, opened 2026-09-04
fix(mcp-instrumentation): record errors and session IDs on MCP session spans

liustve:
## Summary

- record exception details and ERROR status on mcp.session when an exception escapes a stdio, SSE, or streamable HTTP session
- attach the server-issued mcp.session.id to streamable HTTP session spans before they end
- keep handled tool-result errors scoped to their operation spans

## Testing

- full MCP instrumentation test file: 16 passed
- black, isort, flake8, and git diff checks pass


liustve:
Addressed the Claude observations in `3cf8c3e`:

- Added dedicated Streamable HTTP session-error coverage using the existing real MCP client/server test flow. It asserts the session span is `ERROR`, validates the actual `McpError`, and verifies the same server-generated `mcp.session.id` on the client, server, and session spans.
- Kept `CancelledError` outside `except Exception` intentionally so cancellation remains control flow rather than being reported as an MCP session failure.
- Made error and session-ID telemetry recording fail-open so instrumentation failures cannot replace the application exception or prevent normal session-span cleanup.

Local MCP suite: 17 tests and 17 subtests passed.

wangzlei (dismissed):

wangzlei (approved):

PR #882, opened 2026-09-04
fix(code-correlation): treat all site-packages files as third-party

haneric00:
Fixes `code_attributes_test` failing consistently (`41 != 408`): client spans were capturing the OpenTelemetry instrumentation wrapper's line number instead of the user's call site. Root cause is a namespace-package collision — every `opentelemetry-instrumentation-*` distribution collapses to one mapping key, so `is_user_code()` resolves the frame to an arbitrary sibling (`opentelemetry-instrumentation-exceptions`) that isn't in `3rd.txt` and gets misclassified as user code. Fix adds a site-packages backstop: any file under `purelib`/`platlib` is never treated as user code, regardless of metadata/allowlist resolution. Verified the captured line flips from the wrapper frame back to the user call site, and the existing `test_packages_resolver` suite still passes.


wangzlei (approved):

liustve:
closing: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/883

PR #885, opened 2026-09-08
fix(test): pin langchain ChatOpenAI test to Chat Completions API

srprash:
## Problem

The `unit-tests` CI jobs (Python 3.10-3.14) are failing repo-wide, including on `main` via the scheduled "Run Instrumentation Tests" workflow (e.g. run 34262108982). This is **not** caused by any feature PR — it is an upstream breaking change picked up by the unpinned `langchain-latest` tox env.

The failing test:

```
aws-opentelemetry-distro/tests/amazon/opentelemetry/distro/instrumentation/langchain/test_langchain_instrumentor.py::TestLangChainInstrumentor::test_llm_attributes_langchain_models
```

fails with:

```
TypeError: Responses.create() got an unexpected keyword argument 'presence_penalty'
.../site-packages/openai/_legacy_response.py:369: TypeError
```

## Root cause

A recent `langchain-openai` release changed `ChatOpenAI._generate` to route through the OpenAI **Responses API** (`self.root_client.responses.with_raw_response.create(...)`) by default instead of the Chat Completions API. The Responses API does **not** accept `presence_penalty` / `frequency_penalty`.

The test constructs `ChatOpenAI(..., presence_penalty=0.3, frequency_penalty=0.5, ...).invoke(messages)`, which now raises inside langchain/openai before any span is produced.

## Fix

The test's intent is to verify that GenAI request attributes are captured — including `gen_ai.request.frequency_penalty` and `gen_ai.request.presence_penalty` — which are only valid on the Chat Completions API. Pass `use_responses_api=False` to the `ChatOpenAI(...)` constructor so it stays on Chat Completions and the penalty parameters remain valid. This keeps the test exercising exactly the request attributes it asserts.

This is a test-only change scoped to the single langchain test. `use_responses_api` is a supported `ChatOpenAI` field (confirmed present in the resolved `langchain-openai 1.6.1`).

## Testing

Reproduced and verified locally in a Python 3.11 venv with the `langchain/requirements.latest.txt` deps (`langchain-openai 1.6.1`, `openai 3.9.0`):

- Without the change: the test fails with the exact CI error `TypeError: Responses.create() got an unexpected keyword argument 'presence_penalty'`.
- With the change: `test_llm_attributes_langchain_models` passes (`1 passed`).


liustve:
related: https://github.com/langchain-ai/langchain/releases/tag/langchain-openai%3D%3D1.6.1
I think we should not edit the behavior of the tests but support both legacy and new changes

liustve (approved):

wangzlei (approved):

PR #887, opened 2026-09-08
Gh actions workflow optimization

haneric00:
Apply changes listed here:

- Drop labeled and unlabeled from the types: list. The full five-matrix build currently re-fires on every label change on any PR.
- Add a concurrency group (wiki section 1). There is none today, so each new push leaves the prior run's jobs queued beside it.
- Gate the heavy jobs on draft status and add ready_for_review to types: (section 2). Nothing in the file references draft today, so the full suite runs on every draft push.

Addressing ticket cut to us by github admin team to reduce congestion on github actions.


.github/workflows/pr-build.yml
```diff
@@ -5,8 +5,12 @@ on:
       - opened
       - reopened
       - synchronize
+      # labeled/unlabeled kept so adding the "Skip Changelog" label re-runs the
+      # changelog check. Heavy jobs are gated off for these actions (see per-job
+      # `if:` below) so a label change does not re-fire the full build matrix.
       - labeled
       - unlabeled
+      - ready_for_review
```
srprash, line 13: what is this for? Today I think the CI runs for draft PRs as well. Is this for limiting the CI to ready PRs only?

srprash: NVM, saw the comments below

srprash (approved):

PR #889, opened 2026-09-09
fix(langchain): propagate first input, last output, and system instructions to internal agent spans

liustve:
## Problem

LangChain agent spans can start before middleware or graph nodes materialize the actual prompt. In that flow, `on_chain_start` has no input or system instructions to record, while later child LLM callbacks receive the resolved messages and output.

This ordering is common with dynamic prompts, memory or state transforms, and supervisor/worker handoffs. The previous span processor could copy LLM-client attributes, but it could not populate the agent content.

## Changes

- collect the first non-empty input and system instructions plus the latest output from nested LLM callbacks
- apply that content to the owning internal `invoke_agent` span before it ends
- isolate collected content for each nested agent
- remove the separate LangChain span processor
- intentionally keep LLM-client attributes on child chat/text-completion spans instead of copying them to internal agent spans, including provider, request/response model, temperature, `top_p`, token usage, and other per-call attributes
- add a real supervisor with four worker agents whose inputs are materialized after their agent spans start
- validate exact first input, last output, system instructions, parent relationships, and shared trace context for every agent
- update LangChain contract expectations and the `#889` changelog entry

## Testing

- clean Python 3.14 LangChain tox suite: 43 passed, 34 subtests passed
- focused lazy-input and LLM-attribute tests: 3 passed
- `black --check` on changed Python files
- `git diff --check`


liustve:
@claude If `gen_ai.request.temperature` is expected on `invoke_agent` spans, what about `gen_ai.request.top_p` and the other LLM request attributes? Should all request attributes be propagated from the child LLM span, or should these remain scoped to the LLM span?

liustve:
The removal is intentional. LangChain `invoke_agent` spans are in-process orchestration spans, so LLM-client attributes such as provider, request/response model, temperature, `top_p`, token usage, and other per-call values remain on the child chat/text-completion spans. This PR now propagates only the first input, last output, and system instructions to the owning agent span. The unit and contract tests have been updated to reflect that boundary.

wangzlei (approved):

PR #894, opened 2026-09-14
Reuse main build artifacts and images in release build

haneric00:
Images and artifacts we use to release are currently rebuilds of the same versions already validated in a previous main build. Why don't we just use those already built images that are already testing by main build in release? 

We replace the build step in release-build with a promotion step that pulls the main build images from ECR and promotes them to the release build environment.


.github/workflows/release-build.yml
```diff
@@ -97,6 +101,20 @@ jobs:
...
+      # Promotion source: assume the role that main-build uses to push to the staging ECR, so we can
+      # pull the validated image back. NOTE: this needs pull (ecr:BatchGetImage /
+      # GetDownloadUrlForLayer) on the staging repo -- push perms alone are not enough.
+      - name: Configure AWS credentials for staging ECR (image source)
+        uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838 #v5.0.0
+        with:
+          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
+          aws-region: ${{ env.AWS_STAGING_ECR_REGION }}
+
+      - name: Log in to staging ECR
```
v-iw, line 137: Currently: This replaces the release-time image rebuild with promotion of the staging image identified by <package-version>-<release-HEAD-short-sha>. This is the right direction because we want to release an image that already passed main-build validation.

Gap: The wheel and Lambda layer are selected using a specific main-build run ID, but the image is selected independently from the release checkout’s HEAD. If the branch advances, they may come from different commits. A rerun can also reuse this tag, so it may not identify the exact image validated by the selected run.

Request: Could main build record the pushed multi-arch image digest, upload it as metadata with the other artifacts, and have release promote repository@sha256:<digest> from that same selected run? This would guarantee that the released image is the exact artifact that passed validation.

v-iw (dismissed):
Thanks, I reviewed the rehearsal run. It successfully built the real amd64+arm64 image, recorded the immutable index digest, promoted that digest without rebuilding, copied both platform and attestation manifests, and confirmed the source and destination index digests match. Approve.

.github/workflows/release-build.yml
```diff
@@ -149,16 +191,30 @@ jobs:
...
         with:
-          push: true
-          context: .
-          file: ./Dockerfile
-          platforms: linux/amd64,linux/arm64
-          tags: |
-            ${{ env.RELEASE_PUBLIC_REPOSITORY }}:v${{ env.VERSION }}
-            ${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ env.VERSION }}
+          name: staging-image
+          path: staging-image
```
srprash: We should move this step above the "Publish to PyPI" so that all the steps that actually push an artifact to public channel are the last ones to execute. This minimizes partial published states.

.github/workflows/release-build.yml
```diff
@@ -55,6 +62,21 @@ jobs:
         run: |
           gh run download "$RUN_ID" -n release-artifacts -D dist
           gh run download "$RUN_ID" -n layer.zip -D layer
+          gh run download "$RUN_ID" -n staging-image -D staging-ref
+
+      # Guard against a mis-typed version: the wheel built by the selected run must carry the exact
+      # version we're releasing, before anything is published/tagged.
+      - name: Assert wheel version matches release version
```
srprash, line 69: Should we do similar version guard for the image too?

.github/workflows/release-build.yml
```diff
@@ -74,6 +96,12 @@ jobs:
           name: layer.zip
           path: layer/${{ env.LAYER_ARTIFACT_NAME }}
 
+      - name: Upload staging image reference
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
```
srprash, line 100: Looks like we are uploading twice? Already uploaded at line 127

.github/workflows/release-build.yml
haneric00: Should be implicitly covered as we use the hash value of the commit used to run release build to get the main build artifacts (link main build artifacts to the commit version of the release build)

.github/workflows/release-build.yml
haneric00: currently don't see the duplicate, therefore should be good

.github/workflows/release-build.yml
```diff
@@ -131,6 +180,15 @@ jobs:
...
+      # actually pushes to a public channel -- PyPI, then the ECR promote -- runs last and back-to-back,
+      # minimizing the window for a partial-published state.
+      - name: Download validated image reference
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 #v5.0.0
+        with:
+          name: staging-image
+          path: staging-image
+
       # The step below publishes to TestPyPI in order to catch any issues
       - name: Publish to TestPyPI
```
v-iw: Here is the PyPI publishes first, with skip-existing: true.

.github/workflows/release-build.yml
```diff
@@ -149,16 +207,21 @@ jobs:
...
-        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 #v6.18.0
-        with:
-          push: true
-          context: .
-          file: ./Dockerfile
-          platforms: linux/amd64,linux/arm64
-          tags: |
-            ${{ env.RELEASE_PUBLIC_REPOSITORY }}:v${{ env.VERSION }}
-            ${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ env.VERSION }}
+      - name: Promote validated main-build image to release ECRs
```
v-iw: [P1] Image promotion still runs after the PyPI publish

Following up on @srprash's earlier comment: moving the download earlier catches a missing staging-image.txt, but the image itself is still first accessed after PyPI publishes. If imagetools create fails (for example, the staging digest was removed, the role can't pull, or ECR throttles), vX.Y.Z is already published to PyPI with no matching image. Re-running recovers from temporary errors because of skip-existing: true, but not if the staging image no longer exists.

Could we either:

1. add a preflight step before Publish to TestPyPI that runs docker buildx imagetools inspect "$(cat staging-image/staging-image.txt)", or
2. move the promotion step before the PyPI publishes? An image tag can be overwritten, but a PyPI version can't be reused.
Option 1 is the smallest change. Option 2 closes the gap more completely.

Also I think your mock run may not cover this risk.

PR #896, opened 2026-09-15
Add dependency-edge derived attributes to span-metrics allowlist

srprash:
## Description

Extends the span-metrics derived-attribute allowlist (`SpanMetricsConnector`) with the additional OTel semantic-convention attributes needed to derive dependency-edge metrics. All are copied if present via the existing flat `_copy` allowlist — no family branching, no synthesis.

| Family | Attributes |
|---|---|
| Messaging | `messaging.operation.type`, `messaging.consumer.group.name` |
| Peer ([server](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/)) | `server.address` (string), `server.port` (int) |
| GenAI | `gen_ai.request.model`, `gen_ai.provider.name`, `gen_ai.operation.name` |
| AWS resource identity | `aws.s3.bucket`, `aws.dynamodb.table_names` (string array), `aws.lambda.invoked_arn`, `aws.sns.topic.arn`, `aws.sqs.queue.url` |
| FaaS | `faas.invoked_name`, `faas.invoked_provider`, `faas.invoked_region`, `faas.trigger` |

Types match the semconv attribute types: `server.port` stays numeric, `aws.dynamodb.table_names` stays a string array, the rest are strings.

## Testing

- **Unit tests** cover all 16 new attributes deterministically (per-family, plus a not-allowlisted-key-dropped negative test). Suite passes against both the newest and oldest pinned OpenTelemetry versions (89 passed each).
- **Contract-style test** drives a real HTTP request through the requests instrumentor and asserts `server.address`/`server.port` flow onto the metric datapoint. The remaining families (GenAI, FaaS, AWS resource identity) are unit-only because real instrumentation cannot produce those spans in the harness.
- black / isort / flake8 / pylint clean.

## Follow-ups (not addressed here)

Surfaced in review, to revisit separately:
- `aws.dynamodb.table_names` is emitted as an array-valued datapoint dimension; confirm downstream metric-backend compatibility with array dimensions.
- `server.address` is potentially high-cardinality.
- The peer contract test toggles a process-global semantic-convention stability flag and touches an OpenTelemetry-internal symbol; fine sequentially, but worth hardening for parallel test runs.


wangzlei (dismissed):

srprash:
Overriding and merging since the CI failures are unrelated to this change.

PR #898, opened 2026-09-18
fix(llamaindex): exclude broken workflows 2.24.0

liustve:
## Summary

- exclude `llama-index-workflows==2.24.0` from LlamaIndex auto-instrumentation compatibility
- apply the same exclusion to latest unit-test and contract-test dependency resolution
- align the contract-test `llama-index-core` floor with the instrumentor-supported `>=0.13.0,<1` range

## Context

`llama-index-workflows==2.24.0` breaks `FunctionAgent` and `ReActAgent` runs before model or tool execution with `TypeError: unhashable type: FunctionAgent`. The failure is independent of instrumentation and prevents expected tool and downstream spans from being produced.

This follows the exclusion added by OpenTelemetry in https://github.com/open-telemetry/opentelemetry-python-genai/pull/739. The lower workflows bound remains `1.0.1` because ADOT supports `llama-index-core` 0.13.x, which depends on workflows 1.x.

## Testing

- `106 passed, 7 subtests passed` in the Python 3.12 latest LlamaIndex unit suite
- Black check passed for touched Python files
- unit-test and contract-test requirements both resolve `llama-index-workflows==2.23.3`
- direct `Context(FunctionAgent)` smoke test succeeds with the resolved dependency set


liustve:
TODO: add a changelog for this once post release workflow is run

wangzlei (approved):

